PR #37, opened 2016-07-25
Clock takover edge to clock mode

### LINKED ISSUE #31: Rising edge of clock
The LS013B7DH03 requires SPI clock mode 'second rising', whereas other displays use 'first falling'. 

I presume the field `sck_takeover_edge` in `u8x8_display_info_t` can be used for this, but is there a standardized mapping of values to clock modes?

I can create a PR to fix this, e.g. taking the modes as defined by https://en.wikipedia.org/wiki/Serial_Peripheral_Interface_Bus#Mode_numbers?

(Yes, the LS013B7DH03 driver has the wrong clock mode in the display info structure, but I did not notice this until I refactored my platform driver, because I hard coded the mode.)

basilfx:
This PR should fix #31. Can you give it a look, and see if you like it?

I don't have any Arduino boards here, but I think it is still compatible. Any chance you can test this?

For my SSD1306 connected via RIOT-OS, it works. Will test the other display tomorrow.


olikraus:
So you renamed sck_takeover_edge to sck_clock_mode, right?
There two access functions to the renamed variable:
u8x8_GetSckTakeoverEdge(u8x8)
u8x8_GetSckPolarity(u8x8)

basilfx:
I renamed `sck_takeover_edge` to `sck_clock_mode`, and extended its definitions. `sck_takeover_edge` would only describe the clock phase, whereas `sck_clock_mode` also defines the clock polarity. So two bits are used.

If I reasoned correctly, `sck_takeover_edge = 1` implies data is output on rising edge. This corresponds to [clock mode 1](https://en.wikipedia.org/wiki/Serial_Peripheral_Interface_Bus#Mode_numbers).

The macro `u8x8_GetSckTakeoverEdge` simply takes `sck_clock_mode & 0x1`, so you'll get the 'original' value of `sck_takeover_edge`. I don't have an Arduino, but it should map on SPI_MODE0 - SPI_MODE3 eventually (see the third parameter of spi.beginTransaction on https://www.arduino.cc/en/Reference/SPI).

olikraus:
No doubt, i definitly like this extention.

basilfx:
I found an Arduino, so I'll test the SPI changes using a SSD1306.

One thing I might change is the name of `sck_clock_mode`. Not its actually mean 'serial clock _ clock _ mode' :-)

basilfx:
Updated the PR. I have tested this for HW and SW SPI using SSD1306, and it works. 

Note that I have only adopted SPI mode for HW, using `SPI.setDataMode(u8g2->display_info->spi_mode)` (which was `SPI.setDataMode(SPI_MODE0)`). For software, a similar change is needed that inverts the clock high/low for modes 2 or 3. Except for the LS013B7DH03, it will currently not affect other displays.

I have changed `sck_clock_mode` to `spi_mode` to better reflect its purpose. I also changed `u8x8_GetSckTakeoverEdge` to `u8x8_GetSPIClockPhase` and `u8x8_GetSckClockPolarity` to `u8x8_GetSPIClockPolarity`.

Please check it out, and review it :) I'll squash/merge all commits to one single commit if you give the green light.

olikraus:
> Now it is actually means 'serial clock _ clock _ mode', which is double
good finding... renaming makes sense.

olikraus:
ok, i will do the merge. However i probably need to add some comments for myself after the merge.

Thanks a lot for your work here. Looks much more consistent now.

I think SW SPI is less critical here. Due to the timing (data and clock edges do not happen in parallel) the clock polarity does not matter (... probably...)

olikraus:
I think one bug is there
Instead
`#define u8x8_GetSPIClockPolarity(u8x8) (((u8x8)->display_info->spi_mode & 0x10) >> 1)`
it should be:
`#define u8x8_GetSPIClockPolarity(u8x8) (((u8x8)->display_info->spi_mode & 0x02) >> 1)`
right?

olikraus:
There was also a syntax error in your patch in U8x8lib.cpp. I fixed this also (actually it was a naming mistake from my side...)

basilfx:
> I think one bug is there
> Instead
> #define u8x8_GetSPIClockPolarity(u8x8) (((u8x8)->display_info->spi_mode & 0x10) >> 1)
> it should be:
> #define u8x8_GetSPIClockPolarity(u8x8) (((u8x8)->display_info->spi_mode & 0x02) >> 1)
> right?

Correct. My bad. The only use will be the one for SW SPI :-)

PR #47, opened 2016-08-16
Add support for MINI12864 (UC1701)

pdenes:
Ported the driver from u8glib, almost the same as DOGS102 (UC1701), just
with a different size. Tested with a display salvaged from a cheap ebay
MP3 player.


olikraus:
Nice. Thanks a lot. However i guess 
/* flipmode_x_offset = */ 30,
is wrong. Did you test any of the HW flip modes?

pdenes:
You're absolutely right, well spotted! I've tested the flip modes now and also found that the two modes were reversed: the init sequence is setting 0x0a0, 0x0c8 -- which was flip mode 1 originally, so I'm swapping them to make this (the defaul) mode 0. Looks like 'dogs102' has the same inconsistency.

PR #49, opened 2016-08-17
Fix flip modes for MINI12864

pdenes:
- Swap original 'flip0' and 'flip1' to be consistent with the default
  settings
- Set flipmode_x_offset to 4


olikraus:
Thanks!

PR #69, opened 2016-10-08
Preliminary Microchip XC8 compiler support

basilfx:
I know that this library is designed for use with GCC, but that did not stop me from trying to get this to work on a Microchip MCU using the XC8 compiler (free version). To be honest, it doesn't work (yet), but I think the following fixes are more or less general fixes.

I'll open another issue for the remaining problems.

### DIFF
--- a/csrc/u8g2_polygon.c
+++ b/csrc/u8g2_polygon.c
@@ -70,7 +70,7 @@ static uint8_t pge_Next(struct pg_edge_struct *pge) PG_NOINLINE;
 static uint8_t pg_inc(pg_struct *pg, uint8_t i) PG_NOINLINE;
 static uint8_t pg_dec(pg_struct *pg, uint8_t i) PG_NOINLINE;
 static void pg_expand_min_y(pg_struct *pg, pg_word_t min_y, uint8_t pge_idx) PG_NOINLINE;
-static void pg_line_init(pg_struct * const pg, uint8_t pge_index) PG_NOINLINE;
+static void pg_line_init(pg_struct *pg, uint8_t pge_index) PG_NOINLINE;
 
 /*===========================================*/
 /* line draw algorithm */
--- a/csrc/u8x8_d_st7565.c
+++ b/csrc/u8x8_d_st7565.c
@@ -143,7 +143,7 @@ uint8_t u8x8_d_st7565_common(u8x8_t *u8x8, uint8_t msg, uint8_t arg_int, void *a
 	controller: It is not allowed to write beyond the display limits.
 	This is in fact an issue within flip mode.
       */
-      if ( c + x > 132u )
+      if ( ((uint16_t) (c + x)) > 132u )
       {
 	c = 132u;
 	c -= x;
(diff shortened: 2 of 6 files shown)


csrc/u8g2_polygon.c
```diff
@@ -70,7 +70,7 @@ static uint8_t pge_Next(struct pg_edge_struct *pge) PG_NOINLINE;
 static uint8_t pg_inc(pg_struct *pg, uint8_t i) PG_NOINLINE;
 static uint8_t pg_dec(pg_struct *pg, uint8_t i) PG_NOINLINE;
 static void pg_expand_min_y(pg_struct *pg, pg_word_t min_y, uint8_t pge_idx) PG_NOINLINE;
-static void pg_line_init(pg_struct * const pg, uint8_t pge_index) PG_NOINLINE;
+static void pg_line_init(pg_struct *pg, uint8_t pge_index) PG_NOINLINE;
```
olikraus: Why shell we remove "const" here??? This is fine and legal C.

csrc/u8x8_d_st7565.c
```diff
@@ -143,7 +143,7 @@ uint8_t u8x8_d_st7565_common(u8x8_t *u8x8, uint8_t msg, uint8_t arg_int, void *a
 	controller: It is not allowed to write beyond the display limits.
 	This is in fact an issue within flip mode.
       */
-      if ( c + x > 132u )
+      if ( ((uint16_t) (c + x)) > 132u )
```
olikraus: This point is more complicated. 
It is true and intended that there may happen an implicit conversion.
I do not know why there should be a warning. 
I have studied user manual for the XC8, but i do not know why the warning is there and how it should be fixed. (personal note: sizeof(int)==2 for XC8 according to chapter 5.4.2)
I would pefer a separate issue to improve this (if required).

olikraus:
Thanks for providing the pull request and please excause that i refuse also this pull requires. see my comments for each change.

csrc/u8g2_polygon.c
basilfx: I got this error:

```
u8g2_polygon.c:248: error: (984) type redeclared
u8g2_polygon.c:248: error: (1098) conflicting declarations for variable "pg_line_init" (csrc/u8g2_polygon.c:247)
```

Because `const` was missing in the actual implementation. I removed const from the prototype because the methods above also don't have `const`.

csrc/u8x8_d_st7565.c
basilfx: As far as I understand (according to [this](http://stackoverflow.com/a/17882460)), `c + x` is promoted to `signed int` (because all the values of `uint8_t` can be expressed as `signed int`). Therefore, the warning is right, because `132u` is of type `unsigned int`. So to silence the warning, it should be casted to `unsigned int`, or remove the `u` modifier from 132.

I still don't understand why I cannot reproduce this with GCC.

basilfx:
No problem. Discussing the issues is also valuable. Just want to help where possible :-)

PR #71, opened 2016-10-08
Add support for SH1107

basilfx:
I got this OLED display from Visionox, and it seems it is very similar to the SSD1306 (and probably the SH1106).

Someone else is working to get it up and running, but it would be useful if the skeleton is already in the repository.

It's an 128x128 OLED. Datasheet is [here](https://www.dropbox.com/s/nwuj3s6iuajvb9a/SH1107V1.0.pdf?dl=0).


olikraus:
The problem is: by updating codebuild also the online dokumentation gets updated. This may confuse users. So codebuild should be updated along with the actual code. 
I would be happy to include this, but please do this together with the actual code or comment out the section in codebuild

will refuse this for now.

PR #155, opened 2017-01-12
add support for arduino 101

coloz:
add support for arduino 101(arc)


coloz:
I have verified that the hard SPI and hard IIC screens are available on arduino 101

olikraus:
Thanks!

PR #220, opened 2017-03-15
Bitmaps: add bitmap mode

3DevoHendry:
This new mode work in the same way as the existing font mode: It allows drawing
solid (both white and black pixels in the bitmap affect the output) or
transparent (only white pixels in the bitmap affect the output). Transparent
bitmaps may be used to draw multiple bitmaps over each other, for example when
drawing animations. As a side effect, transparent XOR mode now also works by
setting the bitmap mode to transparent and the draw color to 2.

Before this commit, the bitmaps XBM bitmaps were being drawn as solid
object, while the regular bitmaps were drawn as transparent.

The structure of the code is based on how `u8g2_font_decode_len()` handles the
same thing when drawing glyphs.

### DIFF
--- a/cppsrc/U8g2lib.h
+++ b/cppsrc/U8g2lib.h
@@ -185,6 +185,8 @@ class U8G2 : public Print
       { u8g2_DrawLine(&u8g2, x1, y1, x2, y2); }
 
     /* u8g2_bitmap.c */
+    void setBitmapMode(uint8_t is_transparent) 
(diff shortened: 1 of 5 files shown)


3DevoHendry:
Hey olikraus, I have added some examples like you requested.

olikraus:
wow... amazing.

olikraus:
thanks, i have added the required documentation

PR #287, opened 2017-06-06
Contrast command wrong

Tinky1233:
The command for the contrast should be 0xC1h according to datasheet

Now works but poor range, maybe other driving parameters can be set to allow better range.


olikraus:
Excellent finding. It is correct in the init sequence, but wrong in the set contrast function...

PR #327, opened 2017-08-08
[fix] misspelling in u8x8_d_sbn1661.c

legath:
Just for build from master branch


olikraus:
Yes sure... I was working on this code yesterday, but did not do any compile...

PR #346, opened 2017-09-03
Declared "weak" (only for GCC) the function u8x8_GetMenuEvent

gnlcosta:
In this way anyone has the necessity  to customize the input device for userInterfaceSelectionList function (or each other function) he is not obligated ad modify the library.
The "weak" attribute can be used also for other compiler, each compiler has its syntax for "weak".
I used it to define in my code the function u8x8_GetMenuEvent which manages an encoder with push button as input interface.

This patch (method) doesn't change in any way the library and its functionalities. It give only more possibility to the user (for now only to the GCC user).

Ciao


olikraus:
Good idea...

Issue #340, opened 2017-08-23
Improve render-to-file (e.g. tga) support


matthijskooijman:
I was previously using u8glib with its pbm backend to generate screenshots from my application for a manual. With u8g2, the pbm backend was removed, but I did find a tga backend that could fulfill the same usecase.

However, the current tga backend in `sys/tga/common` seems to be very much tailored to the images generated for the u8g2 manual, where the 128x64 LCD screen is rendered at a hardcoded x3 size in the middle, with some room around it for descriptions. This is implemented using two separate u8g2 devices that both write to the same memory. It took me a while to figure out how these things fit together, and now I know I've found that this makes the current tga code unsuitable for what I need (create 128x64 and 128x(more than 64) screenshots without any annotation). I see a few ways to improve this, and I'm curious to your opinion:
 - I added a third device to `sys/tga/common` which implements just a plain tga renderer, without any scaling or other things I do not need. This seems to work, but introduces some duplicate code.
 - I just found that `tools/font/build` also has a tga rendering backend, very similar to the one I already added (except that it does not allow changing the size, but that should be easy to add).
 - I can imagine a cleaner way to unify these three tga backends would be to introduce a "dummy" device, which only renders to memory (essentially just making `DRAW_TILE` a noop). This memory can then be accessed directly by some TGA helper functions, which export the memory contents to a TGA file. To also support the u8g2-manual usecase, this TGA export would need simple compositing support, e.g. allow assembling the exported TGA from two dummy device memories: an LCD part and a (transparent) description part. This compositing needs to happen outside of the normal drawing routines (so we cannot draw the lcd in a smaller memory device and then draw the resulting LCD memory onto a bigger description device), since the resulting image should contain more than 2 colors.
  Writing this, it seems that this compositing stuff might become complicated quickly, so it might be better to use an existing library for doing that (or probably even for doing all the description drawing). A quick search comes up with [CImg](http://cimg.eu/), a single-file C++ library that can handle drawing and writing out images in various formats.

Does any of these approaches seem reasonable to you? For our own manual, I will probably explore the third option, since that would probably also give me options for highlighting parts of my screenshots.

olikraus:
I never though that someone else could be interessted in these tga export devices.
What is your purpose/application/product here?

It would not be too complicated to create a custom tga device for you only, if you specify your needs (variable size, bg/fg color selectable, etc)
[...]

PR #351, opened 2017-09-08
Improvements to the new bitmap code

matthijskooijman:
This relates to issue #340 and contains the improvements discussed in that issue.


cppsrc/U8x8lib.cpp
```diff
@@ -67,6 +67,7 @@ size_t U8X8::write(uint8_t v)
 /*=============================================*/
 /* callbacks */
 
+#ifdef U8X8_USE_PINS
```
olikraus: What exactly is the purpose of this #ifdef statement?
Did you test the lib with and without U8X8_USE_PINS define?

matthijskooijman: The `u8x8_gpio_and_delay_arduino` function, which is guarded by this `#ifdef` uses other features (such as https://github.com/matthijskooijman/u8g2/blob/4f926cba3722017f7447a7a76649adb8dd4956b5/csrc/u8x8.h#L312 and https://github.com/matthijskooijman/u8g2/blob/4f926cba3722017f7447a7a76649adb8dd4956b5/csrc/u8x8.h#L343) which are already guarded by `U8X8_USE_PINS`. So without this added guard, when compiling without `U8X8_USE_PINS`, compilation would fail because these things were missing.

I tested this on a regular machine without `U8X8_USE_PINS`. I think I did not test on an Arduino with `U8X8_USE_PINS` defined, but in theory nothing should change when this define is present.

olikraus: yes, true

PR #387, opened 2017-10-14
Use SPI block transfer on the ESP8266

dratini0:
On the benchmark, I think this it gave around 5 fps improvement on my ST7920 screen, but I didn't note down my results. I can re-do them, if necessary.


olikraus:
No further response? Basically I can not merge because of the mentioned doubts.

dratini0:
Sorry, what do you mean? What doubts do you have?

olikraus:
The risk of breaking the releasefor the ESP8266  because of the 32 bit alignment are to high from my perspective, so i will reject this pull request.

dratini0:
Uhh, I have discovered that it breaks as soon as I link it with GDBstub. I am not sure if it is related, but it is broken nonetheless. I will investigate. Thank you!

PR #390, opened 2017-10-15
Added Multiple SPI Display FAQ

neilpanchal:
Add FAQ item for multiple SPI displays


olikraus:
Thanks!

akamustang:
I was reading the FAQ and noticed both displays have `/*RS*/ 4` which had me confused as it contradicts the statement that separate RST lines are required.

Also I think it's worth mentioning the ability to do a global reset on a single pin via the `U8X8_PIN_NONE` method, although a bit hacky it seems like good information to have in the FAQ.

akamustang:
Wow, crazy fast response, thanks for replying, it reads much clearer now.

olikraus:
Thanks

PR #411, opened 2017-10-30
ls013b7dh03: wrap initialization in transaction.

basilfx:
Since RIOT-OS changed the SPI API, I noticed that my LS013B7DH03 did not work anymore. That's because I forgot to/did not wrap the initialization in a transaction.

This PR fixes it, and makes my screen work again.


olikraus:
makes sense...

PR #416, opened 2017-11-06
Fix compilation without U8X8_USE_PINS

HendryKaak:
The ifdef guards were removed in merge commit '2bcaeb8'. Breaking
compilation outside of the Arduino environment.


matthijskooijman:
Note that these guards were originally added as part of #351.

olikraus:
position of the endif seems to be better now :-)

PR #456, opened 2017-12-12
Bytes are traditionally 8 bits long, at least on my st7920

dren-dk:
Hi, I have no idea where the 9 bits per byte code came from, but it certainly didn't work with my st7920.

After I changed the 9 to an 8 it started working correctly.


olikraus:
You have to use the 4-wire interface for the ST7920. The 3-wire interface requires indeed 9 bit. 
The 3-wire 9-bit interface is NOT supported by the ST7920, but it is supported by more modern controller, so changing this will break many other displays.

PR #494, opened 2018-01-27
add component.mk to work with esp-idf seamlessly

liebman:
By adding a simple component.mk file u8g2 can be loaded as a submodule into an esp-idf projects components directory.


olikraus:
ok...

Issue #418, opened 2017-11-11
Seeedstudio Grove - OLED Display 1.12" not working/SH1107


joaodafonseca:
Hi,

I'm trying to run the examples of the library with the Seeedstudio OLED Display 1.12" but I only get glichy characters on the screen! 

I'm using the following constructor:
U8X8_SSD1327_SEEED_96X96_SW_I2C u8x8(/* clock=*/ SCL, /* data=*/ SDA, /* reset=*/ U8X8_PIN_NONE);  // Seeedstudio Grove OLED 96x96

and using the pins 18 and 19 for SDA and SCL.

Any ideas? thanks in advance
João

joaodafonseca:
Looks like they use a new chip on the modules, SH1107G.
Any change that it's compatible with this library?

olikraus:
> Any change that it's compatible with this library?

Is there a link to the SH1107G controller datasheet?

joaodafonseca:
here: 
https://github.com/SeeedDocument/Grove_OLED_1.12/raw/master/resources/SSD1327_datasheet.pdf

thanks for looking into it!
J

olikraus:
Unfortunately this is NOT the sh1107 Datasheet.

joaodafonseca:
hmmm, found it in the oled wiki page:
http://wiki.seeed.cc/Grove-OLED_Display_1.12inch/

olikraus:
Yes, but for the any update I need to know how to program the SH1107. This is not mentioned in any documents of the seeed home page.

The SH1107 datasheet should be provided from "sino wealth". It will contain the command table of the SH1107 controller.

olikraus:
The SH1107 is not even mentioned on the vendor home page. Maybe it is too new.

http://www.sinowealth.com/zh/product.asp?cat_name=OLED&cat_id=35&class_name=OLED%20%C7%FD%B6%AF&class_id=100&from_page=top

joaodafonseca:
The library provided by seeed already works with the SH1107. Not sure if it helps!!!

https://github.com/Seeed-Studio/OLED_Display_96X96

olikraus:
Well, partly... For example this code is very strange:
https://github.com/Seeed-Studio/OLED_Display_96X96/blob/master/SeeedGrayOLED.cpp#L290

Without knowing the datasheet of the controller it is difficult to guess what the software does.

joaodafonseca:
yeah I see...what about this? 
[...]

PR #499, opened 2018-01-29
Update sh1107 128x64 to be 64x128 instead

micooke:
Regarding issue #418


olikraus:
wow... but i guess I need to change also the codebuild script:
https://github.com/olikraus/u8g2/blob/master/tools/codebuild/codebuild.c#L132

micooke:
What a pain hey :confused:
Sorry, i can add that in as well if you like

olikraus:
Great. Yes, codebuild is the script which generated most of the .cpp code. Also setup and memory management code is generated from there.

PR #517, opened 2018-02-20
Convert to SDL2.

basilfx:
SDL 1.2 appears to be broken on macOS, so I converted U8g2 to SDL2.

schermafbeelding 2018-02-20 om 23 04 43


olikraus:
ok, let's see

olikraus:
ok, looks good. Thanks for the porting. I guess I am sometimes too lazy to move to new versions...

PR #523, opened 2018-02-26
Support for Winstar 48x64 I2C display

matteoscordino:


olikraus:
well, ok, but there is more todo.
- u8x8.h has to be updated
- We probably should add the other interfaces also
- My previous code should be removed from the 64x48 file
- The template files (.ino) for the u8g2/u8x8 Arduino example have to be modified: (A) the 64x48 to 48x64 change and (B) the interface change from SPI to I2C (https://github.com/olikraus/u8g2/tree/master/tools/inoupdate)

Can you add some of the points from above to this pull request?
I can do the rest...

matteoscordino:
I think I fixed everything, but I can't check any of the Arduino stuff.
The only thing I'm not confident about is the interface change from SPI to I2C: the template files were already referencing I2C, not SPI

olikraus:
> already referencing I2C, not SPI

ah, ok... I did not remember this

Let me check everything. Thanks for the update.

olikraus:
looks good, only a small issue with u8x8.h. Thanks for the PR.

Issue #568, opened 2018-04-10
Support for 64x32 SH1106


micooke:
Could you please create a profile for a 64x32 OLED with SH1106 controller, HW SPI.

I will try and get a datasheet for you, but it is for the OLED inside many of the cheap NRF bluetooth smart watches i.e. ID107, DO003.

Unfortunately the SSD1306 does not work for this instance, I get noise for the bottom half and the top half of the screen is clipped (I will post a photo).

olikraus:
I have created beta 2.22.11, download (as usual) from here: https://github.com/olikraus/U8g2_Arduino

The constructor is not yet part of the examples, but all constructor variants exists, example:
```
U8G2_SH1106_64X32_1_4W_HW_SPI u8g2(U8G2_R0, /* cs=*/ 10, /* dc=*/ 9, /* reset=*/ 8);
```

I took over code from the sh1106 72x40 device, but I just updated the dimensions.

micooke:
Tested, a few minor changes were needed. Please see PR #570

olikraus:
> Tested, a few minor changes were needed.

Obviously ;-) 
I did only minimal changes. Thanks for the PR!

You need another beta release?

micooke:
Please :+1:

micooke:
No rush :blush:

olikraus:
beta 2.22.12 is available: https://github.com/olikraus/U8g2_Arduino

olikraus:
I think I can close this...

micooke:
Ah yep, sorry i forgot :no_mouth:

PR #570, opened 2018-04-11
issue #568

micooke:
* update multiplex ratio (was still set to 40px height), y display offsets and x display offsets


micooke:
For reference, this is the .ino i used for the test. It includes text, a bounding box around the screen and Up/Down flipping.
https://gist.github.com/micooke/ea42f8be3f642d1bed5f5b731efff15d

PR #620, opened 2018-05-22
fix unicode start pos bug

acevest:


olikraus:
This is not a bug. You need to update the U8g2 code instead.

acevest:
ok, thank you.

PR #623, opened 2018-05-24
Add sys/avr/at90usb1286

fornellas:
This is an implementation of u8x8_byte_4wire_hw_spi for AVR uCs. It was tested with a at90usb1286, but should work for most AVRs, with minimal changes (possibly only to u8x8_gpio_and_delay).

I'm porting my project https://all3dp.com/best-3d-slicer-software-3d-printer/ from u8g to u8g2, and figuring out all of this was a bit of a pain. This should help other users.

While on the subject, how about putting this u8x8_byte_4wire_hw_spi at u8g2/csrc/u8x8_byte.c? It seems u8g2 is too much Arduino focused, we could use some better C support.

### DIFF
--- a/sys/avr/at90usb1286/main.c
+++ b/sys/avr/at90usb1286/main.c
@@ -1,0 +1,159 @@
+#include <u8g2.h>
+#include <util/delay.h>
+
+u8g2_t u8g2;
+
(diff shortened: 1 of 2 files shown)


fornellas:
Fair point. I'll cut a new pull request, to make this AVR code as a lib at sys/avr/, so that other users can reuse it without copy-pasting code.

olikraus:
Didn't I already merged your pull request?

PR #716, opened 2018-10-03
Set page (row) address correctly for st75256_jlx256160

AnthonyDiGirolamo:
I hooked up one of these displays [purchased here](https://www.aliexpress.com/item/20PIN-SPI-COG-12864-LCD-Module-with-Chinese-Font-ST7567R-Drive-IC-3-3V-White-Backlight/32608104073.html). 

Using the MIRROR rotation worked but produced garbage data at the top.

screenshot_2018-10-02_19-34-52

This pull request sets the y offset correctly for default and flip mode.

screenshot_2018-10-02_19-35-23

Here is a [photo album of all rotations](https://photos.app.goo.gl/baackYN8VwBMCJfN7). Images are:
before fix R0, R1, R2, R3, MIRROR, MIRROR + setFlipMode(1)
after fix R0, R1, R2, R3, MIRROR, MIRROR + setFlipMode(1)

rotation_images

Arduino sketch code used:

```c++
#include <Arduino.h>

#define U8X8_HAVE_HW_SPI
#define U8G2_16BIT

#include <SPI.h>

#include <U8g2lib.h>

// Printing with stream operator
template<class T> inline Print& operator <<(Print &obj,     T arg) { obj.print(arg);    return obj; }
template<>        inline Print& operator <<(Print &obj, float arg) { obj.print(arg, 4); return obj; }

static const uint16_t gamma_table_2_5[] = {
	/* gamma = 2.5 */
	0,	0,	0,	1,	2,	4,	6,	8,
	11,	15,	20,	25,	31,	38,	46,	55,
	65,	75,	87,	99,	113,	128,	143,	160,
	178,	197,	218,	239,	262,	286,	311,	338,
	366,	395,	425,	457,	491,	526,	562,	599,
	639,	679,	722,	765,	811,	857,	906,	956,
	1007,	1061,	1116,	1172,	1231,	1291,	1352,	1416,
	1481,	1548,	1617,	1688,	1760,	1834,	1910,	1988,
	2068,	2150,	2233,	2319,	2407,	2496,	2587,	2681,
	2776,	2874,	2973,	3075,	3178,	3284,	3391,	3501,
	3613,	3727,	3843,	3961,	4082,	4204,	4329,	4456,
	4585,	4716,	4850,	4986,	5124,	5264,	5407,	5552,
	5699,	5849,	6001,	6155,	6311,	6470,	6632,	6795,
	6962,	7130,	7301,	7475,	7650,	7829,	8009,	8193,
[...]

PR #731, opened 2018-10-29
fix jlx256160 flip mode sequences

### LINKED ISSUE #717: Fix Mirror Function for JLX256160
Reference for the original Implementation: https://github.com/olikraus/u8g2/issues/561

Pull Request for this topic. https://github.com/olikraus/u8g2/pull/716

AnthonyDiGirolamo:
Followup to #716 and fixes #717. Thanks for the pointers!
Here are the images using the same arduino code as shown in https://github.com/olikraus/u8g2/pull/716#issuecomment-433728745

screenshot_2018-10-29_09-20-02
screenshot_2018-10-29_09-20-19

Data scan direction (0xBC) is set to 0x00 for flip0 and 0x03 for flip1. This makes sense from looking at the [datasheet](https://www.crystalfontz.com/controllers/Sitronix/ST75256/409/).

screenshot_2018-10-29_09-26-28

### DIFF
--- a/csrc/u8x8_d_st75256.c
+++ b/csrc/u8x8_d_st75256.c
@@ -129,6 +129,29 @@ static const uint8_t u8x8_d_st75256_jlx172104_flip1_seq[] = {
   U8X8_END()             			/* end of sequence */
 };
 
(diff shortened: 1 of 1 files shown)


olikraus:
ok, thanks... looks good.

Issue #754, opened 2018-11-27
Why using binary constants instead of HEX?


NailAgliev:
https://github.com/olikraus/u8g2/blob/d7a20ce5f89c1729721b917025599c57ca105c55/csrc/u8x8_d_st7586s_erc240160.c#L195-L203

Hi all. My IDE(Keil) not support that definition of binary constant. Why just not to use HEX?

olikraus:
True. Code was contributed. But yes, needs to be fixed.

olikraus:
Thanks

PR #755, opened 2018-11-28
#754 change bin to hex

NailAgliev:
That was all places were need changes.


olikraus:
Thanks (yes... i know i just started with the first bin value)

PR #798, opened 2019-01-29
Allow externally allocated buffers

### LINKED ISSUE #796: Allow each display to have its own buffer
I am driving multiple displays from the same Arduino-compatible board and would like to have a way to have a separate buffer for each display. 

The reason is that each display shows a dynamically generated animation with only part of the picture changing in each re-draw. In the current situation, making updates to one display also overwrites what's displayed on the other one.

Current alternatives are:

1. Clear the display buffer every time - this means I'll need to re-create the previous frame of the animation, which is computationally expensive
2. Keep a copy of the buffer for each display in my code - will result in always having 1 more buffer than needed, and unnecessary back-and-forth copying of memory
3. Solution previously discussed in #540, but is a temporary work-around that requires manual modification of the library code

Possible fixes: 

1. In display setup constructor, set a flag that you want the display to allocate its own memory instead of sharing a buffer. This would get passed all the way to the functions in `u8g2_d_memory.c`, who could return a dynamically allocated buffer instead of a static one. Drawback: a lot of code updates to add this to every constructor
2. Implement a `U8G2::setBufferPtr(uint8_t *buf)` that allows the user to allocate their own buffer and pass it to the library, essentially changing the pointer stored internally. 

I can try to implement the second solution and open a pull request unless you see any potential drawbacks/side effects?

futureshape:
Added support for library users to allocate their own buffer and ask the library to use this instead of the built-in statically allocated buffer. This allows users to optionally have a separate buffer per display.

Fixes #796 with the preferred solution discussed in this issue. 

Note: Because a new method has been added, you also need to update:
* `keywords.txt` in `U8g2_Arduino` repositiory
* wiki docs

### DIFF
--- a/cppsrc/U8g2lib.h
+++ b/cppsrc/U8g2lib.h
@@ -166,6 +166,7 @@ class U8G2 : public Print
     uint8_t nextPage(void) { return u8g2_NextPage(&u8g2); }
     
     uint8_t *getBufferPtr(void) { return u8g2_GetBufferPtr(&u8g2); }
+    void setBufferPtr(uint8_t *buf) { u8g2_SetBufferPtr(&u8g2, buf); }
     uint8_t getBufferTileHeight(void) { return u8g2_GetBufferTileHeight(&u8g2); }
     uint8_t getBufferTileWidth(void) { return u8g2_GetBufferTileWidth(&u8g2); }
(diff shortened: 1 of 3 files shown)


olikraus:
as mentioned, I will not include this.

PR #809, opened 2019-02-08
Refactor avr libs.

fornellas:
Cleanup a bit of the mess at the Avr lib, making it more reusable across projects.

The AVR HW SPI function (u8x8_byte_avr_hw_spi) was improved:

- Add support for externally defining the ports for SCK and MOSI externally.
- Properly configure the SPI bus CPHA / CPOL as defined by u8x8->display_info->spi_mode.
- The SPI clock is configured based on u8x8->display_info->sck_clock_hz.

I also organized 2 examples for 2 popular Avr and added a README to help people get started easier.


olikraus:
Thanks, but this will delete all the pictures from here: https://github.com/olikraus/u8g2/wiki/u8g2as7
This does not make sense to me. Is it intended to remove this wiki page?

If not, please provide a patch, which keeps the pictures intact.

fornellas:
Oh, I totally missed that wiki page, sorry. My intention was to remove the old atmega328/main.c, as it has a lot of code duplication with the at90usb1286 example I submitted in the past, but uses software SPI instead of HW SPI.

I never used Atmel Studio, but AFAIK, it does not use avr-libc, that this pull request uses, so code might be incompatible.

How about having:

- sys/avr/avr-libc/
- sys/avr/atmel-studio/

and keep both? Obviously, wiki links would be updated.

olikraus:
I created a new dir (as7) for the png files, merge done...

PR #812, opened 2019-02-10
Separate Atmel Studio and avr-libc

fornellas:
After https://github.com/olikraus/u8g2/pull/809, sys/avr/ became a bit messy, with code for Atmel Studio and avr-libc altogether. Given those are separate libs / environments, let's keep them separated to avoid user's confusion.

I just updated https://github.com/olikraus/u8g2/wiki/Porting-to-new-MCU-platform#atmel-avr to make things clearer.

If merged, a few wiki links will need updating as well.


olikraus:
grmummelgrummelgrummmel

PR #827, opened 2019-02-22
change nextpage loop to sendbuffer

JamesGKent:
found a minor error in the fullbuffer example


olikraus:
well, yes, not a real error, because page loop will work in full buffer mode also.... but yes, good finding.

olikraus:
Thanks!

PR #870, opened 2019-04-15
Updated to fix partial update, add refreshDisplay, for SSD1607_WS_200…

TomWS1:
…x200
Modified UpdateArea.ino included here.  Includes original field update, as well as rotated display, and two new icon fields that are updated every cycle.    These changes are based on code and sequences derived from Waveshare demo code.

[UpdateArea_V2.zip](https://github.com/olikraus/u8g2/files/3081913/UpdateArea_V2.zip)


v1993:
Dear @olikraus, can you, please, don't merge it (at least C++ part) for now? I don't want to change C++ code before adding it into mainstream because of various complications (which will be eleminated after merge).

PR #910, opened 2019-06-03
Add AVR-libc hardware i2c support and example project

QB4-dev:
Hello Oli - thanks for AMAZING library!
I've added AVR-libc implementation of hardware i2c for communication, and an example for atmega328P and SSD1306 128x32 OLED.

In my private repo I made some experiments with ELM-CHAN's xprintf functions and u8g2
https://github.com/QB4-dev/AVR-misc/tree/master/avr-i2c-oled

Maybe it's worth to implement them in u8g2 as standard printf functions for non-Arduino projects?


olikraus:
sure, why not... thanks for the contribution

Issue #991, opened 2019-09-09
STM32F103 Works on SW_SPI but not in HW_SPI


BSG-SalvoL:
I have an STM32F103 board with the STM32duino bootloader. Using U8G2_ST7920_128X64_F_SW_SPI constructor the display works fine, but becouse my project requires more FPS i tried to use U8G2_ST7920_128X64_F_HW_SPI. With this constructor the board freeze with the PC13 led blinking fast and nothing works.
Tried U8G2_ST7920_128X64_F_2ND_HW_SPI too and all other components works fine but not the LCD where nothing is displayed.
The display is connected on SPI2 interface on 
MOSI -> PB15
SCK -> PB13
SS -> PB12
I tried to connect to SPI1 too with no results.
Am i missing something or there is a problem selecting the right SPI?

olikraus:
HW_SPI will use SPI.h of your Board Environment. 
At least for Uno, calling the SPI functions works without problem. So my suggestion is to search for any incompatibilies of the SPI Arduino functions for your environment. 
I just rely on the fact that the same procedures behave in the same way across all boards.

BSG-SalvoL:
Thank you for your response, i've already using the other SPI port (this board has two) with more MCP23S17 and them work fine. 
Due to your response i'm trying to switch the ports for the components. 
It can be a good idea to pass a SPIClass instance to the constructor.

olikraus:
I just use the standard SPI object, which is defined by the Arduino Environment: https://github.com/olikraus/u8g2/blob/master/cppsrc/U8x8lib.cpp#L750

BSG-SalvoL:
Ok maybe i found what's the problem. 
For first for me SPI_INTERFACES_COUNT was not defined, so U8X8_HAVE_2ND_HW_SPI will not be defined by your library; i've actually defined manually in my code. 
Now SPI1, that is used in the library is of type spi_dev* for me which declaration is in file spi_f1.c while SPI according to arduino library is of type SPIClass.

olikraus:
Hmmm

BSG-SalvoL:
I've done another test. 
Defined a pointer in SPI.h and SPI.cpp
`extern SPIClass STMSPI1; //SPI.h
SPIClass STMSPI1(2); //SPI.cpp`

And modified in the function SPI1 with STMSPI1 and still not working.

olikraus:
Will u8g2 Work with SPI instead of SPI1?

[...]

PR #995, opened 2019-09-14
Resolved issue #991. STM32F1 Maple NULL Pointer on SPI2 interface.

BSG-SalvoL:
Signed-off-by: BSG-SalvoL <salvo.graci95@gmail.com>

### DIFF
--- a/cppsrc/U8x8lib.cpp
+++ b/cppsrc/U8x8lib.cpp
@@ -45,6 +45,15 @@
 #include <Wire.h>
 #endif
 
+/* ISSUE #991 */
+#if defined(STM32F1) || defined(STM32F1xx) || defined(STM32F4) || defined(STM32F4xx)
+SPIClass U8G2_SPI2(2);
+bool First_Time = true;
+#else
+SPIClass U8G2_SPI2 = SPI2;
+#endif
+
+
 /*=============================================*/
 
 size_t U8X8::write(uint8_t v) 
@@ -814,6 +823,14 @@ extern "C" uint8_t u8x8_byte_arduino_hw_spi(u8x8_t *u8x8, uint8_t msg, uint8_t a
 extern "C" uint8_t u8x8_byte_arduino_2nd_hw_spi(U8X8_UNUSED u8x8_t *u8x8, U8X8_UNUSED uint8_t msg, U8X8_UNUSED uint8_t arg_int, U8X8_UNUSED void *arg_ptr)
 {
 #ifdef U8X8_HAVE_2ND_HW_SPI
+  /* ISSUE #991 */
+  #ifdef SPI_RESET_FUNC 
+  if(First_Time) {
+    U8G2_SPI2.resetSettings(2);
+    First_Time = false;
+  }
+  #endif
+
   uint8_t *data;
   uint8_t internal_spi_mode;
  
@@ -829,7 +846,7 @@ extern "C" uint8_t u8x8_byte_arduino_2nd_hw_spi(U8X8_UNUSED u8x8_t *u8x8, U8X8_U
       data = (uint8_t *)arg_ptr;
       while( arg_int > 0 )
       {
-	SPI1.transfer((uint8_t)*data);
+	U8G2_SPI2.transfer((uint8_t)*data);
 	data++;
 	arg_int--;
       }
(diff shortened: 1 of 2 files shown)


cppsrc/U8x8lib.cpp
```diff
@@ -45,6 +45,15 @@
 #include <Wire.h>
 #endif
 
+/* ISSUE #991 */
+#if defined(STM32F1) || defined(STM32F1xx) || defined(STM32F4) || defined(STM32F4xx)
+SPIClass U8G2_SPI2(2);
+bool First_Time = true;
+#else
+SPIClass U8G2_SPI2 = SPI2;
```
olikraus: Not so nice, this will make a copy of the complete structure, wasting that much memory in RAM.

olikraus: One more point: at this time, we do not know whether SPI2 exists.

cppsrc/U8x8lib.cpp
```diff
@@ -814,6 +823,14 @@ extern "C" uint8_t u8x8_byte_arduino_hw_spi(u8x8_t *u8x8, uint8_t msg, uint8_t a
 extern "C" uint8_t u8x8_byte_arduino_2nd_hw_spi(U8X8_UNUSED u8x8_t *u8x8, U8X8_UNUSED uint8_t msg, U8X8_UNUSED uint8_t arg_int, U8X8_UNUSED void *arg_ptr)
 {
 #ifdef U8X8_HAVE_2ND_HW_SPI
+  /* ISSUE #991 */
```
olikraus: Why is this only required for the 2nd HW SPI?

cppsrc/U8x8lib.cpp
```diff
@@ -814,6 +823,14 @@ extern "C" uint8_t u8x8_byte_arduino_hw_spi(u8x8_t *u8x8, uint8_t msg, uint8_t a
...
+  #ifdef SPI_RESET_FUNC
```
olikraus: Why SPI_RESET_FUNC? The #ifdefs should match, so this should test for STM32...

cppsrc/U8x8lib.h
```diff
@@ -128,6 +128,12 @@
 #endif
 #endif
 
+/*ISSUE #991 */
+#ifdef BOARD_NR_SPI
```
olikraus: Why does the ESP32 Arduino port use BOARD_NR_SPI? It should better use SPI_INTERFACES_COUNT instead.
From my perspective this should be changed in the ESP32 software.

cppsrc/U8x8lib.cpp
olikraus: In fact the second SPI is called SPI1. This will break on almost all none STM32 plattforms. I can not merge this,

olikraus:
Please excause, but I will not merge this. Your patch will solve two issues:
1) Detection of the second SPI
2) Extra reset for the second SPI

The ESP32 Arduino port seems to deviate a lot from the standard Arduino HAL:
a) dectection for the second SPI is a different vaiable
b) second SPI is called SPI2 instead of SPI1, your patch will break probably all other platforms.
c) extra reset required before using this function

It does not make sense to solve this here in u8g2. All three points should be addressed to the ESP32 project.
I might consider such changes to u8g2, however not in the way as the pull request would do: 
There must be no code change to the existing code. Any changes, which are workarounds for a specific platform must be encapsulated in proper #ifdefs

cppsrc/U8x8lib.cpp
BSG-SalvoL: Oooops i've made a mistake obviusly i've to write SPI1 and not SPI2. 
I know that this can take more Memory but in my SPI lib in which i'll open an issue too, SPI1 is not from SPIClass.

cppsrc/U8x8lib.cpp
BSG-SalvoL: I'm opening an issue to the SPI lib. In Arduino_STM32 library. I don't know why but something is broken there and not in your lib. In fact i'm doing only few modifications only to get sure that all works. Maybe we can check for the SPI _currentSettings if is NULL. 
but the spi_dev inside Is private. And no way to setup and this is required to have SPI communications.

cppsrc/U8x8lib.h
BSG-SalvoL: I know, the lib uses this and not SPI_INTERFACES_COUNT. Is the only way i can figure out the problem.

cppsrc/U8x8lib.cpp
olikraus: ok...

olikraus:
I think the best thing is to close this pull request for now. 
I am always happy to fix bugs and improve u8g2, but in this case i still have the impression that this will be a fix in u8g2 which should happen instead in a different lib. 
Additionally, there is a conflict meanwhile...

PR #1076, opened 2019-12-18
add ARDUINO_ARCH_STM32 to take benfit of Wire.begin(sda,scl) configuration

jmchiappa:
STM32 controllers have several I2C peripherals. So user could choose one of them by selcting the right SDA/SCL.

### DIFF
--- a/cppsrc/U8x8lib.cpp
+++ b/cppsrc/U8x8lib.cpp
@@ -1144,7 +1144,7 @@ extern "C" uint8_t u8x8_byte_arduino_hw_i2c(U8X8_UNUSED u8x8_t *u8x8, U8X8_UNUSE
     case U8X8_MSG_BYTE_INIT:
       if ( u8x8->bus_clock == 0 ) 	/* issue 769 */
 	u8x8->bus_clock = u8x8->display_info->i2c_bus_clock_100kHz * 100000UL;
-#if defined(ESP8266) || defined(ARDUINO_ARCH_ESP8266) || defined(ESP_PLATFORM) || defined(ARDUINO_ARCH_ESP32)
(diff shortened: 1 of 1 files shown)


olikraus:
Did you test this?
STM32 does not support the "Wire.begin(sda,scl)" configuration:

https://github.com/stm32duino/Arduino_Core_STM32/blob/d2ebcd532164a2ed2f74f9337bd86fab766ea133/libraries/Wire/src/Wire.h#L87

    void begin(bool generalCall = false);
    void begin(uint8_t, bool generalCall = false);

This means: Your patch can not work.

jmchiappa:
> Did you test this?
> STM32 does not support the "Wire.begin(sda,scl)" configuration:
> 
> https://github.com/stm32duino/Arduino_Core_STM32/blob/d2ebcd532164a2ed2f74f9337bd86fab766ea133/libraries/Wire/src/Wire.h#L87
> 
> ```
> void begin(bool generalCall = false);
> void begin(uint8_t, bool generalCall = false);
> ```
> 
> This means: Your patch can not work and will probably break the STM32 support.

@olikraus ,
Yes, I did. I've tested this patch on my own stm32duino fork. A PR that adds a new "begin" method has been approved recently [https://github.com/stm32duino/Arduino_Core_STM32/pull/838](https://github.com/stm32duino/Arduino_Core_STM32/pull/838) and will be available soon in the 1.8.0.

olikraus:
Maybe there should be an additional check for the lib version. Is this possible?

jmchiappa:
@olikraus ,
STM32duino 1.8.0 has been released with the new begin method [https://github.com/stm32duino/Arduino_Core_STM32/pull/838](https://github.com/stm32duino/Arduino_Core_STM32/pull/838).
As you requested, I've successfully tested this PR with Stm32duino 1.8.0 on :
- Nucleo L476RG (I2C on pin D44 (SCL) and D43 (SDA)
- Nucleo F411RE (I2C on pin D6 (SCL) and D3 (SDA)

PR #1142, opened 2020-04-02
DrawFrame: Draw corners just once, fixes XOR mode.

rudo-thomas:
Hello.

I noticed that with XOR mode i.e. `setDrawColor(2)`, `drawFrame()` does not draw the corners.
This pull request fixes the ToDo there :-)

Cheers.


olikraus:
Nice finding. Thanks.

PR #1154, opened 2020-04-10
Fix incorrect condition to loop decrement in pg_dec.

ConnorTinen:
While helping to port this library into another language ([Electric Imp Squirrel](https://developer.electricimp.com/squirrel/squirrelcrib)), I was running into errors with the u8g2_DrawTriangle function. I traced these back to what I believe is an error in pg_dec. It appears that pg_dec has a condition holdover from pg_inc that causes it to increment negatively instead of looping back to the maximum index. This pull request is my fix, which worked for my lib.

I am relatively new to this library, so I definitely could be off base on this one. Let me know if there are any questions. Thanks for all of your work on this library, it's extremely useful.

### DIFF
--- a/csrc/u8g2_polygon.c
+++ b/csrc/u8g2_polygon.c
@@ -2,7 +2,7 @@
 
   u8g22_polygon.c
 
-*/	
+*/
 
 
 #include "u8g2.h"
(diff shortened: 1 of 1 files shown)


olikraus:
I think your fix is incorrect. The variable i is unsigned, so  "i < 0" will never become true.
Instead the existing core is correct, because if i==0, then i-- will loop over, i becomes 255 and the existing condition will be true (i > cnt).

Do you have any example where your fix will help?

ConnorTinen:
Ah, right you are. Sorry, I missed that. My fix was based on the language I was porting into which doesn’t have unsigned integers, and I didn’t notice it was unsigned here. My apologies, thanks for the explanation!

Issue #889, opened 2019-05-08
request for 64x32 LCD-Switch


thilda-ma:
Hello,
i would love to use the u8g2 lib with my new LCD-Switches.

But i think they are not in support right now.
The switch is called Q5.
Manual can be found here:
[http://www.lcd-solution.com/fiches_techniques/DataSheet-Q5-V0-9%20RdR01.pdf](http://www.lcd-solution.com/fiches_techniques/DataSheet-Q5-V0-9%20RdR01.pdf)
The manual says mode '0x49' is best for text...

Is it possible to add this one to the library?

Thanks in advance
thilda

olikraus:
I try to get one of these LCDs. And yes, should be possible to add this to u8g2.

thilda-ma:
I have 5 of them at my Place. Maybe i can lend you one, so you don't need to buy it.

olikraus:
I already placed a bid on ebay. But even without the device itself, I should be able to add support. I just need to find some time to implement this.

olikraus:
The display reached my lab...
Arduino Example Sources: http://bplagens.bplaced.net/Q5/

olikraus:
made a mistake during soldering... will put this on hold..

sephalon:
Could you please try if the driver for the [E³] SB6432 from pull request #1180 works for you as well? The protocol of these two LCD keys seems to be compatible.

olikraus:
Ok, I reopen this. I need to find some time for further testing of this device.

PR #1180, opened 2020-05-03
Add [E³] SB6432 LCD key driver

sephalon:
Add support for the [[E³] SB6432 LCD key](http://e3-keys.com/products.html).

These switches feature a programmable RGB LED backlight, for which there is no API function available yet; therefore I added one. What do you think about that? Alternatively, we could "abuse" the set contrast function for this purpose if we widen the `value` parameter to 32 bit.

Documentation is yet to be written.


csrc/u8x8.h
```diff
@@ -471,6 +475,15 @@ void u8x8_d_helper_display_init(u8x8_t *u8g2);
 /*  arg_int: 0..255 contrast value */
 #define U8X8_MSG_DISPLAY_SET_CONTRAST 14
 
+/*
+  Name: 	U8X8_MSG_DISPLAY_SET_BACKLIGHT_COLOR
+  Args:	arg_int: color: 24-bit RGB
```
olikraus, line 480: Actually it is passed via arg_ptr. "arg_int" is 8 bit

olikraus:
wow, you did a deep look into u8g2

olikraus:
Another questions: Are these switches related to #889 ?
I mean, is this the same product: http://www.lcd-solution.com/fiches_techniques/DataSheet-Q5-V0-9%20RdR01.pdf ?

sephalon:
> wow, you did a deep look into u8g2

Thanks! So you think that adding a set backlight color API function makes sense? Then I will write up some documentation for that.

sephalon:
> Another questions: Are these switches related to #889 ?
> I mean, is this the same product: http://www.lcd-solution.com/fiches_techniques/DataSheet-Q5-V0-9%20RdR01.pdf ?

Interesting, at least the interface seems to be compatible. However, the Q5-Dot32-S claims to support alternative addressing modes [1] which the [SB6432 documentation](http://e3-keys.com/files/SB6432.pdf) does not mention. I tried these addressing modes on the SB6432 and they do not work. Since there's also subtle differences (e.g. measures) in the datasheets, I would assume for now that they are different, but mostly compatible products. I will try to confirm this with the manufacturer though.

[1] These would actually fit better to the u8g2 addressing model.

olikraus:
Thanks a lot so far. Plan is to test my device with your code. Unfortunately I do not have much time at the moment. Let's see...

sephalon:
You'll find my attempts to implement the alternative addressing mode of the Q5-Dot32-S at sephalon@b806a14a8aca3a9de3c5299d12bb49dd1d6de3e1.

olikraus:
still no progress on my end...

metizn:
+1 to include E3 LCD switches!

metizn:
+1 to include E3 LCD switches!

olikraus:
I somehow stopped working on this. Is this required for someone?

sephalon:
I am still using these LCD keys in my project and would be interested in getting support for them merged upstream.

metizn:
Would be much appreciated if those LCD keys could be included in U8g2

olikraus:
I own only the switches mentioned here: https://github.com/olikraus/u8g2/issues/889

sephalon:
As I mentioned before, my implementation for the [E³] SB6432 should be fully compatible to the Q5-Dot32-S. I would also be happy to help out with testing the [E³] SB6432. If time permits you can try my implementation of the alternative addressing mode of the Q5-Dot32-S at https://github.com/sephalon/u8g2/commit/b806a14a8aca3a9de3c5299d12bb49dd1d6de3e1; this is untested however since I have only the [E³] SB6432, but not the Q5-Dot32-S.

sephalon:
Pinging this, still interested to get this upstream at some point. Maybe there is some time during christmas break ;-)

olikraus:
Thanks for your hard work, but I probably can't merge this. Some of your modifications are part of generated code. They would be overwritten. Other ports need more review and impact analysis from my end, which would require some time (which I do not have at the moment)

Issue #1172, opened 2020-04-27
Template for new driver (support for S1D15E06 LCD)


nopbit:
Hello,

Thanks for this wonderful library :) I would like to add support for a screen which i removed from an old telephone. It is using chip on glass S1D15E06 with 160x100dots resolution(4 gray scale). Display uses 4 pins Serial Input, Serial Clock, Command/Data and CS pin. 

I created a simple test library with arduino and screen is working perfectly with it. Is there any driver template that i can change and add support for this display?

Best Regards
Mustafa

Datasheet for S1D15E06.
https://www.rockbox.org/wiki/pub/Main/DataSheets/epson_s1d15e06.pdf

olikraus:
This device might be a good candidate:
https://github.com/olikraus/u8g2/blob/master/csrc/u8x8_d_uc1608.c#L473

nopbit:
I edited u8x8_d_uc1608_dem240064_init_seq procedure according to my lcd driver. Now i want to learn how to send data to LCD ? My sample code is like below. I am sending data with row & column. I believe i must play with msg type U8X8_MSG_DISPLAY_DRAW_TILE right?

_My sample code:_

    uint8_t row = 0;
    uint8_t col = 0;
    uint8_t data=0;
    
    _lcd_cmd2(LCD_CMD_PAGE_ADDR,1+(row<<1));
    _lcd_cmd2(LCD_CMD_COL_ADDR,(6*col));
    _lcd_cmd1(LCD_CMD_DATA_WRITE);
     data = 0x0F;
     shiftOut(PIN_MOSI, PIN_SCK, MSBFIRST, b);

> https://github.com/olikraus/u8g2/blob/master/csrc/u8x8_d_uc1608.c#L473

olikraus:
There is basically no nied to use your own functions.
Instead there are several low level functions which allow hardware independent transfer to the display. They also include an abstraction layer, so that a display will work with multiple busses (I2C, 8-bit 4-wire SPI, 9-bit 3-wire SPI, parallel, etc).

The functions are listed here:
https://github.com/olikraus/u8g2/blob/9ecb87c4b62e91302006d278fc3498263a6c567d/csrc/u8x8.h#L597-L603

The first call should go to "u8x8_cad_StartTransfer". 
After this use any data transmit calls (u8x8_cad_SendCmd, u8x8_cad_SendArg, ...)
Last call should be "u8x8_cad_EndTransfer". 

"cad" means "command argument data".

[...]

PR #1190, opened 2020-05-14
Added support for S1D15E06 LCD

nopbit:
Thanks for your support. I added support for S1D15E06 with 160x100 pixels. 

Best regards

### DIFF
--- a/cppsrc/U8g2lib.h
+++ b/cppsrc/U8g2lib.h
@@ -17169,6 +17169,94 @@ class U8G2_MAX7219_8X8_F_2ND_4W_HW_SPI : public U8G2 {
   }
 };
 
+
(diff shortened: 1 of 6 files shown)


olikraus:
Ref: #1172

Nice work, but many files are automatically generated. So some other files need to be updated.
I will probably reject this request and add the files manually.

nopbit:
Ok, thanks. Do you have a special script to auto generate the files?

olikraus:
Once you wrote the u8x8 device, add the device name into u8x8.h, then update the initial table in https://github.com/olikraus/u8g2/blob/master/tools/codebuild/codebuild.c
Compile and execute codebuild.c. This will generate all other files.

lt6210925:
https://blog.csdn.net/lt6210925/article/details/119519261

PR #1224, opened 2020-06-14
Set clock polarity in software SPI implementations

sephalon:
Set the clock polarity in software SPI implementations; previously CPOL was always 0.

I have only implemented this as a PoC for Arduino 4 wire software SPI so far, since I am not sure if I understand the differing implementations well enough and lack the required test setups; so please take this with a grain of salt.


olikraus:
Indeed I also have some doubts. You are probably right, that there is something missing, but wouldn't your implementation lead to an additional transition?

Other question is: Which display did cause problems here?

PR #1255, opened 2020-08-16
Allow passing i2c bus and i2c address in from callers on linux-i2c sys

kenrestivo:
Allow passing i2c bus and i2c address in from callers on linux-i2c sys.


olikraus:
closed as suggested

PR #1320, opened 2020-10-26
Add SSD1362 and added updated SH1122 now work with u8x8

theeprawn:
Added:-
setFontScroll(x,y,gap);
setGrey(value);
new display ssd1362
u8g2.begin() & u8x8.begin() constructor added setFontFlip(0) removed remap from initilation for SH1122 and SSD1362, this is to prevent flipping when doing silent resets of processor
parent class identifier added in u8x8 when display initialized. Helps display controler know how to decode tile buffer orientation

Updated:-
updateDisplayArea() for horiztonal stacked pixel displays(like SH1122 and SSD1362)
SH1122 display controller so works correctly with U8x8 and updateDisplayArea
Arduino keywords setFontScroll & setGrey

### DIFF
--- a/cppsrc/U8g2lib.h
+++ b/cppsrc/U8g2lib.h
@@ -110,7 +110,7 @@ class U8G2 : public Print
     uint8_t getMenuEvent(void) { return u8x8_GetMenuEvent(u8g2_GetU8x8(&u8g2)); }
 
     void initDisplay(void) {
-      u8g2_InitDisplay(&u8g2); }
+      u8g2_InitDisplay(&u8g2, u8g2_parent_class); }
       
     void clearDisplay(void) {
       u8g2_ClearDisplay(&u8g2); }
@@ -120,6 +120,9 @@ class U8G2 : public Print
       
     void setFlipMode(uint8_t mode) {
       u8g2_SetFlipMode(&u8g2, mode); }
+
+    void setGrey(uint8_t value) {
+      u8g2_SetGrey(&u8g2, value); }
 
     void setContrast(uint8_t value) {
       u8g2_SetContrast(&u8g2, value); }
@@ -135,6 +138,7 @@ class U8G2 : public Print
       #ifndef U8G2_USE_DYNAMIC_ALLOC
       initDisplay(); 
       clearDisplay(); 
+      setFlipMode(0);
       setPowerSave(0); 
       return 1;
       #else
@@ -251,6 +255,7 @@ class U8G2 : public Print
     void setFont(const uint8_t  *font) {u8g2_SetFont(&u8g2, font); }
     void setFontMode(uint8_t  is_transparent) {u8g2_SetFontMode(&u8g2, is_transparent); }
     void setFontDirection(uint8_t dir) {u8g2_SetFontDirection(&u8g2, dir); }
+    void setFontScroll(int8_t x, int8_t y, uint8_t gap=0) {u8g2_SetFontScroll(&u8g2, x, y, gap); }
 
     int8_t getAscent(void) { return u8g2_GetAscent(&u8g2); }
     int8_t getDescent(void) { return u8g2_GetDescent(&u8g2); }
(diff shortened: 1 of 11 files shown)


olikraus:
Thanks. I will look into this.

olikraus:
Hmm... the pr fails:
```
../../../../csrc/u8x8_display.c:98:45: warning: unused parameter 'parent_class' [-Wunused-parameter]
 void u8x8_InitDisplay(u8x8_t *u8x8, uint8_t parent_class)
                                             ^
../../../../csrc/u8x8_display.c: In function 'u8x8_SetFlipMode':
../../../../csrc/u8x8_display.c:110:49: error: 'parent_class' undeclared (first use in this function)
   u8x8->display_cb(u8x8, U8X8_MSG_DISPLAY_INIT, parent_class, NULL);
                                                 ^
../../../../csrc/u8x8_display.c:110:49: note: each undeclared identifier is reported only once for each function it appears in
../../../../csrc/u8x8_display.c:108:45: warning: unused parameter 'mode' [-Wunused-parameter]
 void u8x8_SetFlipMode(u8x8_t *u8x8, uint8_t mode)
```
                                             ^

olikraus:
I looked a little bit closer to your implementation. 
1) I would prefer to have a common solution for u8x8 and u8g2 so that we do not need to have two separate implementations. This will also increase codesize of the complete lib. Probably codebuild.c needs to be changed
2) I would prefer to keep the existing implementation and add a new constructor for both displays
3) The SSD1362 probably also misses codebuild.c updates.

theeprawn:
1 & 2. yes, probably need to implement at rotate implementation in u8x8 constructor like u8g2 has with its buffer in setup. But I didnt want mess with your code that much, so made code that bolted on - so to speak.
3. I not sure u8g2 builds corrctly as I cant test, partly why I sent(pull) changes to u8g2_Ardunio first. I can and did test that version.
4. it rotates all u8x8 tiles. u8x8 one sends one tile at a time as not use a buffer, but does use a repeat of tile count

theeprawn:
If I have time, I look deeper into u8x8 code, I using u8g2 for projects, but u8x8 is more helpfull for those with more constrained controllers. It certainly goes faster as well. 
I wondering if some like u8g2->ll_hvline = u8g2_ll_hvline_vertical_top_lsb
can be done in u8x8

theeprawn:
ok, been many months since I looked at this.
In u8g2_buffer.c,  u8x8_DrawTile(u8g2_GetU8x8(u8g2), 0, dest_tile_row, w, ptr); is used to display a buffer row, of tile_width display.
In u8x8_8x8.c all draw with single width except for functions:-
u8x8_ClearDisplayWithTile() and u8x8_ClearLine() in u8x8_display.c which use single tile repeated.

theeprawn:
just thinking, needs a draw_hvtile_cb in u8x8 setup which is used by u8x8_8x8 functions to draw. U8g2 calls the useual u8x8_DrawTile() function.
the draw_hvtile would call u8x8_draw_hvtile_horizontal or u8x8_draw_hvtile_vertical guess in u8x8_hvtile.c
one would pass straight to u8x8_DrawTile() function the other would do rotation on buff before passing it on.

theeprawn:
that should make it all consistent and least code changes, easier to maintain.

theeprawn:
I try implement but will take awhile as busy with jobs, customers are very demanded at moment.
whats the ll stand for in u8g2_ll_hvline_horizontal_right_lsb for naming convention ie. u8x8_ll_hvtile_horizontal_right_lsb

theeprawn:
oh, ll is for low level, guess keep that for u8x8_ll_hvtile.c

olikraus:
> I try implement but will take awhile as busy with jobs, customers are very demanded at moment.

Same here

Please do not spent so much time here. I also think the SSD1362 is much smarter than the SH1122, so maybe just keep the SH1122 as it is.

theeprawn:
Job got postpnded today due to rain and me not feeling great. So had a go at implementing hvtile today. Just got it compling no errors, tested on sh1106 working ok so, will run more test with some test code. Only can test on sh1106, ssd1306, sh1122 and ssd1362. It basically addes a callback in u8x8 in constructor for drawtile, which for u8g2 and u8x8 with vertical tiles is straight through for horizontal it passes on a converted temporary tile buff, which can be more than one tile, but as far as I am aware u8x8 only every sends 1 tile at a time.
I have changed u8x8_clearline and u8x8_cleardisplaywithtile to got through drawtile_cb, also updated display_capture to use horizontal buffer when appropriate. When searching code for references I spotted them,
Anyway fingers crossed. I guessing best way if all good is to make another branch and another pull from that? so can see changes

olikraus:
Actually, i reduced working on u8g2 a lot. Although I added another display today (ST7528) I decided not to add any new features. Goal from my end is to have a stable graphics library, which could be a starting point for further development. 
One problem with your PR is, that there are many changes and i would need a lot of time to understand everything. Time, actually, which i do not really have.  

As mentioned earlier, all new devices need to be registered here: https://github.com/olikraus/u8g2/blob/master/tools/codebuild/codebuild.c
In fact many files are automatically generated out of codebuild.c. I can check the another PR but codebuild.c must be supported and anything else than modifing u8x8_d_xxx.c files should be avoided. Moreover the current sh1122 implementation should not be touched, instead a new driver should be created.

I really like your work and the support for the SSD1362, but currently I see more risks than benefits. Together with the fact, that I am not able to check the PR in detail, I currently will not merge the PR into master.

theeprawn:
I agree, that why I wrote setFontScroll as a bolt on same as setFontRotation, if its #define U8G2_WITH_FONT_SCROLLING is commented out in file u8g2.h then all its functional code is removed as if it not exist, just leaving an empty setFontScroll() function.
I really not see it causing any issue, I specifically wrote it in Atmel asm (I use to mainly write assembler) friendly way and tested it, as I not want to cause you problems. I just like to contribute to your awesome work, I like the way you compressed the fonts with your RLE.
Its up to you if not want to use. The display control code I did similar to not effect anything, but then it has to act different for u8x8 and u8g2.

theeprawn:
Scroll function works by:- (most of work done in u8g2_font.c - u8g2_font_decode_len function)
1.scrolling/rotation done before font_rotation vectoring, simplifies and saves alot of not needed duplication of calulations
2.treating top left corner of scroll glyph box as 0,0 using font offsets to align current glyph to 0,0(removes adjustments before returning from u8g2_font_decode_len)
3.width scroll box is current glyph(char) = width(delta) + scroll gap
4.height scroll box is current fonts = ascent(A) - descent(g) + scroll gap
5.Vertical scrolling/rotation is relativly easy, just adjusts glyph ly and uses boundaries and visible draw limits
6.Horizontal scrolling/rotation does same adjusts glyph lx and using boundaries and draw limits, but has to split the drawn line if it reaches edge of scroll or glyph box.
Hope that helps with understand

theeprawn:
Regards display controller, thanks for pointing to codebuild.c,  now understand how your building. No problem if you want keep orginal sh1122 and I can adjust ssd1362, but they will only work with u8g2 then.
The code I wrote adding u8x8_draw_hvtile.c I tested and works fine and brings consistency between the u8g2 and u8x8 for the hv buffer. It only really used when u8x8 is messaging a display controller to draw that has horizontal architecture, but will add a small amount to code & RAM space used.
As you say your not got time to check and after a stable base code for future, which I agree with. I not want to muddle your hardwork, so I will check into other display controllers code, to check for possible issues, and then make a seperate branch and send you a pull so you can check when you have time and decide if usefull or not. No worries if not.

PR #1327, opened 2020-11-02
u8x8_draw_hvtile.c

theeprawn:
tile buffer conversion for u8x8 to use displays with different memory architure, much the same as hvline for u8g2
adds callback hvtile_cb function before u8x8_Drawtile called.
adds extra line of code in codebuilder to u8x8lib.h   ie. u8x8_SetupTile(getU8x8(), u8x8_draw_hvtile_vertical_top_lsb);

also added sh1122_256x64_midas and ssd1362_256x64_raystar display drivers(also in codebuilder)

I checked through all display drivers, several with horizontal architechure dont use do{} while(arg_int>0), but guess could be updated.


theeprawn:
These are the ones I found not using arg_int, which is by u8x8 to cleardisplay, cleardislaywithtile() ..
u8x8_d_a2printer.c
u8x8_d_lc7981.c
u8x8_d_ls013b7dh03.c
u8x8_d_max7219.c
u8x8_d_sed1330.c
u8x8_d_st7586s_erc240160.c
u8x8_d_st7586s_s028hn118a.c
u8x8_d_st7920.c
u8x8_d_t6963.c

theeprawn:
hmm, if you prefer a cleaner pull without extra display driver, let me know. Just looking, they clutter the changes, but they are in big chucks generally.  They add alot to the changes count.

theeprawn:
Updated prefix naming to same as filename to make unique, so no errors with duplicate names with other display drivers(names, functions)

PR #1361, opened 2020-12-17
Removed sole CR from u8g2_fonts.c

epac-tom:
Removed single CR in u8g2_fonts.c
Editor warns that the file has inconsistent line endings.


olikraus:
u8g2_fonts.c is a generated file .
Patch doesn't make sense.

v1993:
Maybe you should patch source file (or generator?) then? There really shouldn't be a stray wrong line ending.

olikraus:
Maybe you can describe little bit more in detail about what is wrong with the line ending?
What do you see? What do you expect?

Please also see here: https://en.wikipedia.org/wiki/Newline#Representation
And please note, that u8g2 is developed under Linux.

epac-tom:
@olikraus the file uses LF throughout but on line 258 there is a single CRLF. Unfortunely github doesn't show that and looks like I've just removed an blank line.
image

In the IDE I'm currently using it complains about inconsistent line endings and fixes the file, meaning I've got a dirty repo - thought this would be a quick clean up to keep the file consistent.

olikraus:
hmm... this could be an issue in the bdf source file.
Maybe I should update bdfconv for this.
Goal: Remove/replace with 32 all ascii codes < 32 in the copyright string of the bdf file.

PR #1408, opened 2021-02-14
Add teensy t3 support fixes #1175

### LINKED ISSUE #1175: Modify library so that Teensy devices can use i2c_t3 instead of Wire
So i had made a project in the past using a Teensy 3.2 and oled connected to 2nd i2c interface with U8g2. Today i wanted to use the library again to do some tests and i found out my sketch wouldn't compile because i was using the latest library forgetting i had to modify it in the past to make it work. 

Basically on Teensy devices if you want to use the 2nd i2c interface Wire1 doesn't work. You need to use i2c_t3 in place of Wire: https://github.com/nox771/i2c_t3

First issue is that Wire.h is included in U8x8lib.cpp
Since you need to use i2c_t3 in place there as well you need to account for that.

Secondly U8X8_HAVE_2ND_HW_I2C doesn't work and i didn't dive deep enough to find exactly why. I see there is already an issue open regarding this: https://github.com/olikraus/u8g2/issues/779
Workaround for this is to go to U8X8_HAVE_HW_I2C and replace all instances of "Wire." to "Wire1."

jogi-k:


jakplugg:
I edited my u8g2 to use this as per your files, It compiles and starts properly. Does it use the usual u8g2.setBusClock(1000000)? that is what i was previously using with wire.h, but now my display locks up after a few seconds, even after i changed it to 100000. The project has some ISRs running (playing audio) and and they are still working after the display locks up.

PR #1470, opened 2021-04-19
Fixing a bug in the vertical mirroring code.

mnutsch:
The vertically mirrored text had an error which caused it to be distorted (although still somewhat legible). This commit addresses a bug which was causing the distortion. Vertically mirrored text is now clearly legible.


olikraus:
Thanks for the contribution

PR #1524, opened 2021-07-02
Added ability to invert Menu Event button logic

MikeEckels:
This is a small enhancement providing the user the ability to invert the logic of buttons used with menu events. This is helpful when a button is active HIGH rather than active LOW like the library currently supports. A simple bool to toggle inverted logic was added to the 'begin()' method when 'U8X8_USE_PINS' is defined.


MikeEckels:
Bumping this PR...Any desire to merge this? Relatively simple addition, but provides some helpful functionality.

olikraus:
Probably I missed this. But actually it is my goal to remove button functionality. Instead MUI will be there which will support external button libraries: https://github.com/olikraus/u8g2/tree/master/sys/arduino/u8g2_page_buffer/MUIInput2BtnBounce2

MUI: https://github.com/olikraus/u8g2/issues/1747

PR #1543, opened 2021-08-03
Add RT-Thread RTOS support

wuhanstudio:
Hi olikraus, 

Previously I ported u8g2 to RT-Thread, which is a burgeoning RTOS that supports 150+ boards.

https://www.rt-thread.io/board.html
https://github.com/olikraus/u8g2/wiki/Porting-to-new-MCU-platform#rt-thread

U8g2 is very popular among RT-Thread developers and it has been downloaded 7000+ times.

https://packages.rt-thread.org/en/search.html?search=u8g2

It would be great if you are willing to incorporate the port for RT-Thread into the upstream repo, so that developers from the RT-Thread community can always receive the latest feature.

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -1,0 +1,19 @@
+from building import *
+import rtconfig
+
+# get current directory
+cwd     = GetCurrentDir()
+# The set of source files associated with this SConscript file.
+src     = Glob('csrc/*.c')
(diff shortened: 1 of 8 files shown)


olikraus:
looks good...

PR #1547, opened 2021-08-04
add rt-thread cpp support

wuhanstudio:
Hi, thanks a lot for accepting rt-thread,

Previously I only added basic port for C programming. This patch adds **C++ support** for rt-thread port. 

To make things compatible, I only updated files in `sys/rt-thread`, thus this patch should have no effect on other ports.

- Add `sys/rt-thread/tools/codebuild.c` that generates C++ headers for rt-thread automatically
- Add autogenerated `U8x8lib.h` and `U8g2lib.h` in `sys/rt-thread/tools/port`
- Add sw/hw spi/i2c examples for C++

https://user-images.githubusercontent.com/15157070/128245289-e6ffeec9-42f8-4ac0-8dbd-26e0193c2238.mp4


wuhanstudio:
The demo in the video uses RT-Thread Studio as the IDE with the following settings:

https://www.rt-thread.io/studio.html

image

image

PR #1562, opened 2021-08-19
added support to drawStr_P to print from flash memory using PSTR()

ColoMAX:
printing from flash memory instead of RAM is not possible with `DrawStr` or other u8g2 function, only with inherented `print`class and using its `__Flashhelper` overload with the `setCursor()`. But It might me more elegent to use  the added `DrawStr_P()` function, following a similair naming scheme as stdio.h. 

Thus it would be used as such:
u8g2.drawStr_P(0, 9, PSTR("Hello World!"));

in contrast to :
u8g2.setCursor(0,9);
u8g2.print(F("hello World!"));


olikraus:
I have actually avoided to introduce too many PROGMEM functions here. 
In fact you can use u8g2.print() for this purpose. The .print() approach also does not require the introduction of another callback (u8x8_ascii_next_pgm).
The other problem is, that you have used the existing PROGMEM macros, instead of the u8x8 layer for it. This will break compatibility with other systems like ARM or ESP. In other words: Your contributed code will not work on ARM and ESP.

So I currently will not merge this:
- Functionality already exists with .print()
- Code seems to be incompatible with ARM and ESP

PR #1591, opened 2021-09-19
Extend UTF8 encoding for complete compatibility

deividAlfa:
**u8x8_utf8_next** is able to decode the entire UTF8 range, but the encoding exceeds the uint16_t storage size.
Usually it won't be a problem, but the unicode size is simply ludicrous.
To not hurt the performance, maybe we can add a switch to use uint16_t when not needing the extra range.
ex. Something like this:
```
// Defined somewhere, ex u8g2.h
#define USE_EXTENDED_UTF8

// Then
#ifdef USE_EXTENDED_UTF8
    typedef uint32_t encoding_t;
#else
    typedef uint16_t encoding_t;
#endif
  ```

Then use "encoding_t" for all these modified functions, like:
```
u8g2_uint_t u8g2_DrawGlyph(u8g2_t *u8g2, u8g2_uint_t x, u8g2_uint_t y, encoding_t encoding)
  ```

### DIFF
--- a/csrc/u8g2.h
+++ b/csrc/u8g2.h
@@ -1502,9 +1502,9 @@ size_t u8g2_GetFontSize(const uint8_t *font_arg);
 void u8g2_SetFont(u8g2_t *u8g2, const uint8_t  *font);
 void u8g2_SetFontMode(u8g2_t *u8g2, uint8_t is_transparent);
 
-uint8_t u8g2_IsGlyph(u8g2_t *u8g2, uint16_t requested_encoding);
-int8_t u8g2_GetGlyphWidth(u8g2_t *u8g2, uint16_t requested_encoding);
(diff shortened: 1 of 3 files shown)


olikraus:
Is there any bug except, that any encoding > 0xffff will not work?

Another point is this: Changing the encoding size to uint32 will not help as long as the font data structures are not changed: U8g2 internal font format only supports 16 bit encoding. It had been a design decision to support only 16 bit for gylph encoding: A balance between memory consumption and number of glyphs.
Moreover bdfconv is powerful enough to map unicode plane 1 into base plane, so a work around is there (and is also used).
I do not see the need for the merge...

deividAlfa:
I wasn't drawing anything, just playing, figuring how to decode utf8 for diferent thing.
By pure coincidence, I saw that the encoding was missing a number (was: 12345, decoded:2345).
So I decided to investigate and found that.
If it's designed that way, no worries!

PR #1609, opened 2021-10-04
[fix] rm codebuild.o

wuhanstudio:
When I try to `make clean` under the `codebuild` directory, it returns the error:
```
codebuild.o does not exist
```

Because indeed `codebuild.o` is not generated when compiling use `-o`:

```
$(CC) -Wall -g $(LDFLAGS) $(SRC) -o codebuild
```


olikraus:
Ah, yes, good finding.

PR #1610, opened 2021-10-04
Add arm-linux support

wuhanstudio:
Hi,

This port implements the following features for general `arm-linux` boards:

[●] Hardware I2C
[●] Hardware SPI
[●] Software I2C
[●] Software SPI
[●] 8080 mode
[●] Add C++ wrapper
[●] Improve software GPIO speed by tracking fd

Previously this port is in a seperate repo:

https://github.com/wuhanstudio/u8g2-arm-linux

Then several developers think the port has implemented enough features for their projects, thus they suggest that I may upstream the port so that they can receive the latest updates for new devices on arm-linux.

https://github.com/wuhanstudio/u8g2-arm-linux/issues/5:

All the port for arm-linux resides in `sys/arm-linux` folder, so that this port should have no effect on other ports and the functionality of u8g2.

Examles for the arm-linux port can be compiled using:

```
$ cd u8g2/sys/arm-linux
$ make
```


olikraus:
Does it make sense to have U8g2lib.h and U8x8lib.h in the new sys/arm-linux folder? These files are automatically updated and will be outdated soon.

olikraus:
well, then, I go ahead with the merge...

PR #1614, opened 2021-10-05
Use yield() only when #defiend ARDUINO

wuhanstudio:
Hi, 

Some platforms do not provide the yield() funtion.

I'm not sure if this yield() is required, and I uncommented this line for the RTT and arm-linux port. If we do not use yield(), different ports can share the same `U8g2lib.cpp`.

Or it's better to do it this way, if this is required for the Arduino platform:

```
#ifdef ARDUINO
  yield(); 
#endif
```

### DIFF
--- a/cppsrc/U8g2lib.cpp
+++ b/cppsrc/U8g2lib.cpp
@@ -44,7 +44,9 @@ static Print *u8g2_print_for_screenshot;
 
 void u8g2_print_callback(const char *s)
 { 
+#ifdef ARDUINO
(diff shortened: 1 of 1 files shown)


olikraus:
It is/was required for Arduino. #define would be required here.

wuhanstudio:
> It is/was required for Arduino. #define would be required here.

Hi, I've changed it to be `#ifdef ARDUINO`.

olikraus:
Thanks

Issue #1638, opened 2021-10-18
ARM Linux: SPI mode on RPI


shlyakpavel:
Hello!
I think it's a Raspberry PI issue, but I have not checked it on other arm linux boards.
I have successfully connected my st7920 display to both Raspberry PI 4 hardware SPI and atmega32 hardware SPI with u8g2. However, unless I add `internal_spi_mode = 0;` on Raspberry here
https://github.com/shlyakpavel/u8g2/blob/3b6af74b565a1dd44734ac5513dc7ebd998281fa/sys/arm-linux/port/u8g2port.c#L322
I get an error:
```
can't set spi mode: Invalid argument
Aborted
```
I think that comes from RPI SPI driver unable to change SPI modes as they are set somewhere in the bootloader.
The SPI mode in the library is correct as everything works fine on my ATmega32 board. The possible solution could be to check if the system supports setting SPI mode form userspace and skipping this step if not (maybe with a warning). 
I'd love to hear your idea on correct implementation/workaround
Have a nice day!
P. S. for those who question about using RPI port instead of arm linux one: the RPI port is out of date and doesn't work. And I think it won't work on boards above RPI3

sgjava:
@wuhanstudio OK, so I added debug info to your SPI 4 wire example and get:
```
u8g2_InitDisplay
CLK:255, DATA:255, CS:255, RST:199, DC:198
SDA:255, SCL:255
SPI Device Mode Set
spi mode: 0
bits per word: 8
max speed: 500000 Hz (500 KHz)
SPI Device Opened
RESET:199 1
RESET:199 0
RESET:199 1
DC:198 0
Buffering Data 1 
AE 
```
Reading documentation I see:
```
The following figures illustrate the recommended power ON and power OFF sequence of SSD1306
Power ON sequence:
1. Power ON VDD
2. After VDD become stable, set RES# pin LOW (logic low) for at least 3us (t1) (4) and then HIGH (logic
high).
3. After set RES# pin LOW (logic low), wait for at least 3us (t2). Then Power ON VCC.
(1)
4. After VCC become stable, send command AFh for display ON. SEG/COM will be ON after 100ms
(tAF).
```
So shouldn't 0xaf be sent instead of 0xae to turn on display? Also, RST pin should be 0, 1 instead of 1, 0, 1?

wuhanstudio:
Hi, @sgjava 

[...]

PR #1650, opened 2021-10-31
[fix][arm-linux][gpio] fopen w+ --> fopen w

wuhanstudio:
After running several tests, indeed as suggested by @sgjava https://github.com/olikraus/u8g2/issues/1638#issuecomment-955608269, fopen with `w` works for both `ubuntu 5.4.0-1045-raspi` and `armbian 5.10.34-sunxi`.

This PR change all the `w+` to `w` in gpio.c.

### DIFF
--- a/sys/arm-linux/drivers/gpio.c
+++ b/sys/arm-linux/drivers/gpio.c
@@ -47,7 +47,7 @@ static int readIntValueFromFile(char* fileName)
 
 int exportGPIOPin(int pin) 
 {
-    FILE* fp = fopen("/sys/class/gpio/export", "w+");
+    FILE* fp = fopen("/sys/class/gpio/export", "w");
     if (fp != NULL) 
     {
         fprintf(fp, "%d", pin);
@@ -63,7 +63,7 @@ int exportGPIOPin(int pin)
 
 int unexportGPIOPin(int pin) 
 {
-    FILE* fp = fopen("/sys/class/gpio/unexport", "w+");
+    FILE* fp = fopen("/sys/class/gpio/unexport", "w");
     if (fp != NULL) 
     {
         fprintf(fp, "%d", pin);
@@ -95,7 +95,7 @@ int setGPIOValue(int pin, int value)
     else
     {
         GPIO_FILENAME_DEFINE(pin, "value")
-        fp_gpio[pin] = fopen(fileName, "w+");
+        fp_gpio[pin] = fopen(fileName, "w");
         if (fp_gpio[pin] != NULL) 
         {
             fprintf(fp_gpio[pin], "%d", value);


sgjava:
@wuhanstudio take a look at https://github.com/sgjava/u8g2/tree/master/sys/arm-linux as I refactored drivers to use c-periphery. There should be a decent performance improvement overall. Can you pull this down and test it before I make a PR?

wuhanstudio:
Hi, I tried you c-periphery version of u8g2 port on OrangePi One (`armbian: Linux orangepione 5.10.34-sunxi`).

Both tests use `u8g2_4wire_sw_spi`. The c-periphery takes longer time to print out `Done` and exit the program.

Do you obseve similar effects on your system?

sysfs:
```
$ time sudo ../../../bin/u8g2_4wire_sw_spi
Initialized ...

real    0m1.636s
user    0m0.149s
sys     0m0.287s
```

c-periphery:
```
$ time sudo ../../../bin/u8g2_4wire_sw_spi
Initialized ...
Done

real    0m6.653s
user    0m0.077s
sys     0m0.370s
```

sgjava:
You see I put in a 5000 ms delay before shutting display off (i.e. Done)? So in my graphics test I2C is about the same which is not surprising. SPI is a lot faster because you were transferring one byte at a time. I transfer a full buffer each time (up to 128 bytes). Software bit banging should be faster too I suspect.

wuhanstudio:
Turns out there is a 5000ms lol. Sorry I ran the test in a hasty. I'll test the HW SPI soon.

> @wuhanstudio You see I put in a 5000 ms delay before shutting display off (i.e. Done)? So in my graphics test I2C is about the same which is not surprising. SPI is a lot faster because you were transferring one byte at a time. I transfer a full buffer each time (up to 128 bytes). Software bit banging should be faster too I suspect.

sgjava:
@wuhanstudio no worries, I know for a fact SPI is a lot faster based on sendBuffer. If you look at https://github.com/sgjava/java-periphery#gpio-performance-using-perf you can see gpiod on a NanoPi Duo is 242 Khz. Software bit banging should be a lot faster. You need to run a series of sendBuffer.

wuhanstudio:
I encountered an error while using c-periphery on `Armbian Orangepi One 5.10.34-sunxi`:

```
gpio_open(): pin 10, c-periphery library built without character device GPIO support.
gpio_open(): pin 20, c-periphery library built without character device GPIO support.
```

I used `PA10` and `PA20` for RES and DC respectively.

```
#define OLED_SPI_PIN_RES            10
#define OLED_SPI_PIN_DC             20
```

image

sgjava:
Make sure you do a `make clean` in case old artifacts exist. Then try build again. I had this happen to me when I forgot to do a clean in between testing sysfs and gpiod.

wuhanstudio:
Cheers, it works on my device and it's averagely 20% faster than sysfs. Loooking forward to see your PR.

sgjava:
@wuhanstudio So SPI was about the same speed as I2C before. It's a lot faster now. I'll benchmark at some point.

PR #1661, opened 2021-11-04
Replace drivers in arm-linux to use c-periphery

sgjava:
This PR will allow better overall performance and the ability to use newer GPIOD or SYSFS to access gpio pins. @wuhanstudio has tested my fork.


olikraus:
Anything else to merge?

sgjava:
@olikraus no sir, not at this time.

PR #1662, opened 2021-11-04
Small changes

sgjava:
Removed static global variables. Really no reason to keep them static. Also small change to readme pointing to this project instead of my fork. @wuhanstudio

### DIFF
--- a/sys/arm-linux/README.md
+++ b/sys/arm-linux/README.md
@@ -6,7 +6,7 @@
 
 ## Download U8g2 project
 * `cd ~/`
-* `git clone --depth 1 https://github.com/sgjava/u8g2.git`
+* `git clone --depth 1 https://github.com/olikraus/u8g2`
 
 ## Modify source as needed
 Currently GPIO device, I2C and SPI buses are hard coded in u8g2port.c.
--- a/sys/arm-linux/port/u8g2port.c
+++ b/sys/arm-linux/port/u8g2port.c
@@ -7,16 +7,16 @@
 #include "u8g2port.h"
 
 // c-periphery i2c handle
-static i2c_t *i2c_device;
-static const char i2c_bus[] = "/dev/i2c-0";
+i2c_t *i2c_device;
+const char i2c_bus[] = "/dev/i2c-0";
 
 // c-periphery spi handle
-static spi_t *spi_device;
-static const char spi_bus[] = "/dev/spidev1.0";
+spi_t *spi_device;
+const char spi_bus[] = "/dev/spidev1.0";
 
 // TODO: This should be 1 to 1 with pins
 #if PERIPHERY_GPIO_CDEV_SUPPORT
-static const char gpio_device[] = "/dev/gpiochip0";
+const char gpio_device[] = "/dev/gpiochip0";
 #endif
 
 // c-periphery GPIO pins


sgjava:
Just caught this @olikraus , sorry.

wuhanstudio:
Hi, @sgjava

Perhaps it's better to keep them static so that they are only visible within this file, and this also make sure that no one changes these variables unintentionally elsethere.

sgjava:
Well the thought process is to be able to change these, so they are not hard coded. Then the calling program can set them. Should you really have to change the source of U8g2 to set buses? I mean I have a lot of boards and some don't have the hard coded buses. @wuhanstudio

wuhanstudio:
> Well the thought process is to be able to change these, so they are not hard coded. Then the calling program can set them. Should you really have to change the source of U8g2 to set buses? I mean I have a lot of boards and some don't have the hard coded buses. @wuhanstudio

Actually, I mean it's better to design a function to change it so that we know we're changing the bus for u8g2, but not the bus for other applications.

sgjava:
@wuhanstudio I agree, but the issue is one bus to many devices (I2C addresses for instance). It look like there's a PR sitting out there at least for I2C https://github.com/olikraus/u8g2/pull/1255. The you can have several u8g2_t in use. Then you have to revisit u8x8_byte_arm_linux_hw_i2c as it is not thread safe using statics to persist across calls:
`	static uint8_t buffer[32];
	static uint8_t buf_idx;
`

wuhanstudio:
Indeed, as for now we are unable to create mutiple u8g2 instances for different i2c buses.

Looks like this is an issue that has been pending for a long time. What do you think it's the best way to add support for mutiple i2c and spi buses? @olikraus

sgjava:
@wuhanstudio Well, so phase one was to modernize drivers (GPIO, I2C and SPI) which is complete. The second phase was to create a struct to hold this device info and tie it to u8g2_t somehow, so the callbacks would know which one to operate on. We would just need to move all the state into this struct and make the callbacks stateless (maybe only have a global array with these structs in u8g2port). So the thought process is figure out something unique inside u8g2_t to key off of. I'm not sure why @olikraus didn't merge https://github.com/olikraus/u8g2/pull/1255 other than not to pollute the structs with this info.

olikraus:
I probably missed #1255 
But adding #ifdef UNIX would be good.

olikraus:
Can you give me a summary which PRs are required? I didn't follow your discussion and i also away from my desktop PC. (Typing this from my smartphone)

sgjava:
@olikraus ii would be very helpful if you added an ID var (uchar or whatever) to u8x8_struct with a getter and setter. Then you can track displays by ID and we can handle I2C, SPI, etc. info outside of those structures. https://github.com/olikraus/u8g2/pull/1255 is only for I2C, but we really need a way to handle anything in the various ports. Having a unique ID for each display will allow other structures to manage the state. I'm open to @wuhanstudio or anyone else's input.

Hopefully that makes sense. On Linux based SoCs you can have multiple displays and memory is not as much of a concern.

olikraus:
As I mentioned, my access to GitHub is very limited at the moment (only smartphone).

sgjava:
@olikraus I can do it later tonight and submit a PR with just that simple change. The work on the arm-linux one which will be more complex.

PR #1664, opened 2021-11-06
Add I2C and SPI buses

sgjava:
@olikraus This is what we were discussing in the other PR I closed. By moving I2C and SPI bus numbers into u8x8_struct these can be removed from ports and calling programs can set them. I decided against device ID for now. This should allow you to close https://github.com/olikraus/u8g2/pull/1255 unless you want to cherry pick the display specific stuff. This is more generic.

@wuhanstudio this will allow me to remove hard coded buses from arm-linux. I need to think about how to handle I2C state (buffer) a bit and may add device ID back in. I figure one step at a time.


sgjava:
Actually going to add gpio chip number too. Closing this.

olikraus:
ok, I merged #1666 instead

PR #1666, opened 2021-11-07
Make GPIO, SPI and I2C bus stateful

sgjava:
@olikraus This is what we discussed as far as being able to have bus state in u8x8_t for GPIO device SPI and I2C buses. It adds only 3 bytes to the struct, but allows ports to be more configurable and have multiple displays more easily supported.

@wuhanstudio I reworked arm-linux and it's a lot more flexible now without hard coding the buses. Please take a peek at https://github.com/sgjava/u8g2 fork.

### DIFF
--- a/csrc/u8g2.h
+++ b/csrc/u8g2.h
@@ -415,8 +415,14 @@ void u8g2_ClearDisplay(u8g2_t *u8g2);
 #define u8g2_GetDisplayWidth(u8g2) ((u8g2)->width)
 #define u8g2_GetDrawColor(u8g2) ((u8g2)->draw_color)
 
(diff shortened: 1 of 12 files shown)


olikraus:
ok, but i will add some ifdef's here

olikraus:
I have added a macro for this PR:
https://github.com/olikraus/u8g2/blob/2e7b5be78cf0067043a929ad7dbf6677f2bfc1b3/csrc/u8x8.h#L201-L204

sgjava:
@olikraus thank you for the quick turn around. I was actually thinking about ifdef, so I'm glad you added that.

sgjava:
@olikraus looks like you can close https://github.com/olikraus/u8g2/pull/1255 as well. It has conflicts now anyways :)

PR #1673, opened 2021-11-08
Remove GPIO chip and I2C/SPI buses from u8x8_struct

sgjava:
@olikraus I'm using user_ptr now and I believe I have everything in the port thread safe now. It includes the ability to have multiple buses open at the same time and handle GPIO pins for multiple displays.


sgjava:
@olikraus I just added a commit to initialize the arrays to NULL.

sgjava:
Hahaha, I'll have another after I test the software I2C and SPI. I broke out the init functions for hardware/software and fixed the delay.

PR #1676, opened 2021-11-11
Create init user data functions for I2C/SPI HW/SW and optimize SW I2C

sgjava:
I created individual functions to init user data based on I2C/SPI HW and SW. This includes setting pins, so it only ads one line of code for programs. Improved I2C speed by allowing custom (or no) delay.


sgjava:
Thank you sir!

PR #1677, opened 2021-11-11
SW SPI speed up

sgjava:
@olikraus this is what we were discussing about SW SPI. It was the same issue as SW I2C, so I use the same delay value (or none) from my user data struct. I cannot tell the difference between HW and SW, so I will have to do some actual benchmarks. Thanks for your ideas and help.

### DIFF
--- a/sys/arm-linux/port/u8g2port.c
+++ b/sys/arm-linux/port/u8g2port.c
@@ -248,7 +248,10 @@ uint8_t u8x8_arm_linux_gpio_and_delay(u8x8_t *u8x8, uint8_t msg,
 	(void) arg_ptr; /* suppress unused parameter warning */
 	switch (msg) {
 	case U8X8_MSG_DELAY_NANO:            // delay arg_int * 1 nano second
-		sleep_ns(arg_int);
+		user_data = u8x8_GetUserPtr(u8x8);
+		if (user_data->delay != 0) {
+			sleep_ns(user_data->delay);
+		}
 		break;
 
 	case U8X8_MSG_DELAY_100NANO:        // delay arg_int * 100 nano seconds


sgjava:
@olikraus Thanks for the merge. This will make arm-linux very usable now for concurrency, HW and SW. It looks like D0 and D1 were commented out in port for U8X8_MSG_GPIO_AND_DELAY_INIT. I haven't messed with 8080 protocol. Should I enable these in the port? Is this GPIO only?

olikraus:
> It looks like D0 and D1 were commented out in port for U8X8_MSG_GPIO_AND_DELAY_INIT. I haven't messed with 8080 protocol. Should I enable these in the port? Is this GPIO only?

Not sure what you mean here?

Basically D0 and D1 are identical to the SPI pins, to they are not required as such (the number behind is the same)

https://github.com/olikraus/u8g2/blob/b6faa6571f57ba0cc583add78b0909a539dacca9/csrc/u8x8.h#L300-L303

sgjava:
https://github.com/olikraus/u8g2/blob/49ac418a8b3ffce963755fd1fad0f5b6bfad00ec/sys/arm-linux/port/u8g2port.c#L308 in the port. I didn't write this, but I'm trying to figure out why they were commented out and not implemented.

olikraus:
> In the port. I didn't write this, but I'm trying to figure out why they were commented out and not implemented. So my question is should I allow these to be set?

I also didn't wrote that code. But yes, those cases would be required for 8080, 6800 and SW SPI interfaces.

sgjava:
OK, so I'll allow these to be set.

sgjava:
OK, found out U8X8_MSG_GPIO_D0 and U8X8_MSG_GPIO_D1 are dups of U8X8_MSG_GPIO_SPI_CLOCK and U8X8_PIN_SPI_CLOCK.

olikraus:
:-D

I said so, didn't I
> D0 and D1 are identical to the SPI pins

yes:
https://github.com/olikraus/u8g2/blob/49ac418a8b3ffce963755fd1fad0f5b6bfad00ec/sys/arm-linux/port/u8g2port.c#L354

I kept them together in my own HAL procedures....

sgjava:
Hahaha, you you are correct. I'm still on a learning curve a bit. I'll just remove them from the code so it's not confusing. I've been trying to clean up as I go.

olikraus:
> I'm still on a learning curve a bit.

also my fault, the HAL API could be more structured.

sgjava:
I've been in the arm-linux port weeds so much I haven't spent a lot of time in the core. Just eye balling my Java demo it looks like SW I2C is 2x faster than HW I2C at 100K since I made delay changes. Makes sense if NanoPi Duo can bit bang at 242K (libgpiod) and upstream code is efficient.

PR #1679, opened 2021-11-11
NULL GPIO pin array

sgjava:
Need to NULL out pin array even if not used. Java doesn't like this when calling done_user_data otherwise.


sgjava:
Ah, for some reason C is OK with this but my wrapper Java SIGSEGVs!

sgjava:
Thank you.

PR #1680, opened 2021-11-13
Move common user_data_struct init code into function

sgjava:
Move common user_data_struct init code into function and clean up comments.


sgjava:
Thank you sir.

Issue #1682, opened 2021-11-15
New commits broke things on Linux SPI


shlyakpavel:
Previously, I could use my SPI like thishttps://github.com/shlyakpavel/u8g2/commit/3b6af74b565a1dd44734ac5513dc7ebd998281fa
Now, I need to define DC pin even if my screen does not have it. I cannot setup CS pin correctly as u8g2.initSpiHw just doesn't support setting it right. I have tried to fix these to myself, but I didn't get the screen working :(

shlyakpavel:
https://github.com/olikraus/u8g2/pull/1676 broke it

olikraus:
@sgjava ?

sgjava:
Can you give me an example of code that was working here? Then I can try locally.

shlyakpavel:
> @shlyakpavel Can you give me an example of code that was working here? Then I can try locally.

https://github.com/shlyakpavel/u8g2/blob/3b6af74b565a1dd44734ac5513dc7ebd998281fa/sys/arm-linux/examples/cpp-examples/u8g2_4wire_hw_spi/u8g2_4wire_hw_spi.cpp
This code worked well

sgjava:
I think the CS thing I just need to make it where you can pass a pin in.
`u8x8_SetPin(u8g2_GetU8x8(u8g2), U8X8_PIN_CS, U8X8_PIN_NONE);`

shlyakpavel:
There's is also a problem that I am required to set DC pin that doesn't exist on st7920

sgjava:
@shlyakpavel Yeah, I got that too. I'll make it where you can pass a < 0 value for the pin which will result in U8X8_PIN_NONE. Would that work?

shlyakpavel:
Wow
I have manually passed U8X8_PIN_NONE as OLED_SPI_PIN_DC and correct OLED_SPI_PIN_CS and everything works now. It's a dirty hack, so I won't make it a PR. Just saying this to clarify no other problems are here

shlyakpavel:
> @shlyakpavel Yeah, I got that too. I'll make it where you can pass a < 0 value for the pin which will result in U8X8_PIN_NONE. Would that work?

I don't know

shlyakpavel:
> I'll probably make changes anyways, since this may come up again with other displays.

Sure! Thank you

sgjava:
[...]

PR #1684, opened 2021-11-15
Added RES arg to init_i2c_sw and CS arg to init_spi_hw

sgjava:
This should fix https://github.com/olikraus/u8g2/issues/1682


shlyakpavel:
Why did you break formatting?

sgjava:
@shlyakpavel I actually did formatting, it wasn't formatted before.

sgjava:
@olikraus Thanks!

PR #1685, opened 2021-11-16
Exit on error instead of allowing callbacks to proceed

sgjava:
@olikraus this is what we discussed about return codes from callbacks. Instead of proceeding with callback and seg fault this will print error and exit. Much cleaner solution to error handling.


sgjava:
Thank you.

PR #1696, opened 2021-11-18
Added spi_mode and max_speed to user data

sgjava:
All set @olikraus


sgjava:
Crap need to make past tense.

PR #1697, opened 2021-11-18
Added spi_mode and max_speed to user data

sgjava:
All set @olikraus


sgjava:
Thank you.

shlyakpavel:
Why did you
1) Screw correct way of automatically setting SPI mode (replaced it to be always 0)
2) Screw c++ overloaded function (added another function instead). I think its was a correct cpp-way to do it

sgjava:
You want it the old way using display version? What do you suggest?

shlyakpavel:
> You want it the old way using display version? What do you suggest?

I want it to be old way unless advanced function is called. I just suggested it in my PR that was closed to merge yours

PR #1701, opened 2021-11-18
Multiple displays using threads

sgjava:
@olikraus, so this code finally demonstrates multiple software I2C displays running concurrently. This is what I have been shooting for. arm-linux port is thread safe finally.


sgjava:
IMG_20211118_133140

Still image of example program.

olikraus:
Nice picture :-)

sgjava:
@olikraus here's the youtube video https://www.youtube.com/watch?v=3IENDb5GNok. Will be cool once you merge.

sgjava:
I was thinking about this and maybe dynamic buffer should be internalized in init_user_data in the port. Then the program doesn't have to worry about setting up buffer. Let me think about this a bit. I'll close this for now.

olikraus:
Nice video :-)

PR #1703, opened 2021-11-19
Make arm-linux use dynamic buffer everywhere

sgjava:
@olikraus OK, good thing I though about this after sleeping on it. Once you go dynamic you have to do it everywhere or it breaks things (at least in the port). This is not a big deal since as a side benefit displays are thread safe. Tested HW/SW I2C and SPI displays plus multi threading. Should be all good now.


olikraus:
Can you limit the #ifdef in u8g2 to if defined(__ARM_LINUX__)?

sgjava:
You mean not #if defined(__ARM_LINUX__) || defined(unix) || defined(__unix__) || defined(__unix) just __ARM_LINUX__?

sgjava:
@olikraus look at my latest commit.

sgjava:
@olikraus was there something else you were waiting on besides ifdef change?

sgjava:
Thank you sir!

PR #1708, opened 2021-11-21
Minor CPP refactor plus CPP thread example

sgjava:
Refactored some redundant code in CPP, renamed CPP dirs to be consistent and added CPP multi thread/display example.


sgjava:
Thank you.

PR #1716, opened 2021-12-03
Remove exit on error

sgjava:
@olikraus I removed the exit on error and updated the readme. I was wondering if there's a cleaner way to detect errors from callbacks and let the calling program see them. i.e. u8g2_InitDisplay and other methods return an int error code (0 for no error). I'm not sure where to start. I guess I could handle this by adding some probe functions in the arm-linux port that uses same init args and tries at least to create resource (GPIO pin and I2C/SPI busses).

In any event I don't like the hard exit because you still may want your program to execute without the display. I left in the output to stderr, so at least you'll know why.

### DIFF
--- a/sys/arm-linux/README.md
+++ b/sys/arm-linux/README.md
@@ -5,7 +5,8 @@
 * Overall performance should be better.
 * Run as non-root user.
 * Thread safe and multiple display capable.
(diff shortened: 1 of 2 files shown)


sgjava:
Thank you.

olikraus:
No problem

PR #1752, opened 2022-01-10
Add #ifdefs to make C++ interface usable without Arduino.

PetteriAimonen:
I would like to use the C++ interface but without Arduino dependency.

It seems `#ifdef ARDUINO` is already used in a few places in the code, so this pull request expands it so that the C++ interface can be built without any need for Arduino functions.


olikraus:
ok....

PR #1831, opened 2022-04-05
Add support for GP1287AI and GP1247AI VFD

izilzty:
Hi, I added support for GP1287AI GP1247AI VFDs and tested 1 page buffer and 2 page buffer on Arduino 328P and full buffer on STM32.
If there is no problem, I hope it can be merged with the upstream.
I noticed that some files may be auto-generated, but I'm not sure how to modify them, if you can handle it would be greatly appreciated.

sorry for bad english

gp1287ai_u8g2_logo

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,12 +27,3 @@ The Arduino library U8g2 can be installed from the library manager of the Arduin
 
 [Setup Guide and Reference Manual](https://github.com/olikraus/u8g2/wiki)
 
----
-
-增加了对【GP1287AI】和【GP1247AI】两款VFD显示器的支持。
-
-![GP1287AI Logo](https://github.com/izilzty/u8g2/raw/master/doc/gp1287ai_u8g2_logo.png) 
-
-因为SPI字节顺序问题目前仅支持U8g2，不能支持U8x8（显示为镜像），详细信息可以参考u8x8_d_gp1287ai.c文件内的说明。
-
-如需旧版本的U8g2下载地址如下，可以支持U8x8但不能支持软件SPI。https://docs.qq.com/doc/DTlp1akJUVlFVQW5r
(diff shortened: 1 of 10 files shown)


olikraus:
Thanks for your effort. What I do not understand: Why is there a need to change u8g2_ll_hvline.c 
There should be a solution without this modification.
Is the datasheet available?
Is there a way for me to buy this display?

izilzty:
Thanks for reply.

Just like the Sharp ls013b7dh03, this series of VFDs requires the SPI of LSBFIRST. So I swapped the byte order of the command using SWAP8. 
But unlike ls013b7dh03, this series of VFD hardware is drawn from top to bottom, the existing ``u8g2_ll_hvline_vertical_top_lsb`` draws in the correct direction but in reverse byte order. I need to modify the byte order when drawing.

The data sheet for this series of VFDs has been attached. This datasheet is basically the same as the GP1287AI GP1247AI except for the resolution.
[GP1219AI_D_03.pdf](https://github.com/olikraus/u8g2/files/8422390/GP1219AI_D_03.pdf)

Both VFDs are custom made and disassembled from other equipment. I think they may not be readily available outside of China. But they have lower prices and more stock, so I would like to have code to drive them.
I can help you with any tests if needed.

olikraus:
> I need to modify the byte order when drawing.

It would be much better to do this in the device function instead.

The question here is: Do we really need to spent time into this. You have a working copy of u8g2 for your project and the display is anyway not available in small quantities for DIY projects. So I would just keep everything as it is and those users who really got hands on your displays can use your lib.

izilzty:
You are right. There may not be so many people using it. I will keep it in my library. Thank you for taking the time to deal with it.
I will close the pull requests

olikraus:
🙂

Issue #1413, opened 2021-02-26
Using U8g2 with an already-established I2C Wire/Wire1 Instance?


dstulken:
Hello!   I'm attempting to use U8g2 to drive an SSD1309 OLED display, from the second hardware I2C bus on an ESP32, in the Arduino environment.  My test code works fine when using the SW I2C method, so I know the hardware and wiring are OK.  But the display remains blank when using "2ND_HW".
  
I'm aware of Issue #779, "pin remapping for second I2C interface", and that it was put on hold due to the large effort to implement.  

Rather than re-opening #779 (or reviving the modify-every-constructor solution), I had a thought that it might be possible for U8g2 to instead simply skip the I2C bus initialization entirely, and utilize an already existing (and configured) Wire / Wire1 instance.   My project (and I assume many others) already has a "Wire1.begin(dataPin, clockPin, clockRate)" call at the beginning of the setup function to initialize that bus, which other I2C devices use, so the "extra" U8g2 initialization calls to Wire1.begin() and Wire1.setClock() seemed unnecessary in this case.  

I had intended to propose this as an alternate solution to #779. A global #define could be added that could allow any I2C display (2ND_HW or otherwise) to skip the I2C init and re-use an established Wire instance, sidestepping the labor-intensive mass constructor updates entirely, for all displays at once.  But in my testing, with the separate bus init in my code, and a modified U8g2 library with the Wire1 I2C inits commented out, I still only see a blank display. 

I attempted to dig deeper, but for a project as complex as this, had a bit of trouble tracing how the library works and/or what might be the problem.  Do you see any reason why this approach wouldn't work?  

Thanks in advance.

dstulken:
Ok, I found the issue - it was the missing WIRE_INTERFACES_COUNT definition causing the whole code block to not be executed at all.  I posted some info on that at the end of Issue #961.   

With that out of the way, I went back and did a test with all of the Wire1.begin() and Wire1.setClock() calls commented out, and a separate Wire1.begin(dataPin, clockPin, clockRate) call in my setup() function, and it works just fine.  I then commented out that begin(...) function in my setup() function, and it fails as expected (blank OLED), so using an "external" I2C setup is working.

So with that being the case, what do you think of this concept (a flag to disable I2C bus initialization in U8g2) as a global solution to #779?  It could actually apply to any I2C bus (Wire or Wire1), and would actually scale nicely with no changes needed if any future device supports additional (3+) hardware I2C busses.

[...]

Issue #1134, opened 2020-03-21
U8X8-library interferes with A/D-converter PCF8591 on the same I2C-bus - tricky!


HanksSaloon:
Hi. I have a SSD1306 128X64 display and 4-port A/D-converters (PCF8591) on the same I2C-bus, and they seem to interfere. As soon as something is written to the display, the data read back from the A/D-converters is wrong.

Setup: Arduino Nano, and on the I2C three PCF8591s, an EEPROM AT24C32 and a SSD1306 128X64 (The display is behind a 5V to 3v3 level converter).

Here's the simplified schematic 
[PCF8591_test.pdf](https://github.com/olikraus/u8g2/files/4363891/PCF8591_test.pdf)

The device addresses can be found in the program listing below.

`#include "Wire.h"
#include "U8x8lib.h"
#include "AT24C32_64.h"

#define An0104 (0x48) // 1st 4-port AD / PCF8571
#define An0508 (0x49) // 2nd 4-port AD / PCF8571
#define An0912 (0x4A) // 3rd 4-port AD / PCF8571

#define ExtMem (0x56)  // AT24C32 EEPROM
#define LCDaddr (0x78)

U8X8_SSD1306_128X64_NONAME_HW_I2C LCD(/* reset=*/ U8X8_PIN_NONE); // Create LCD
AT24C32_64 EXM(ExtMem); // Create external EEPROM

const int AnSet[3] = {An0104, An0508, An0912 };
byte rv0, rv1, rv2, rv3;
float volt[4];

void setup() {
  Wire.setClock(100000);
  Wire.begin();
  Serial.begin(115200);
  LCD.setI2CAddress(LCDaddr);
  LCD.begin();
  EXM.initialize();  // OPTION 2

  LCD.setFont(u8x8_font_saikyosansbold8_u); // Select font
}

void loop() {
  for (byte i=0; i<=2; i++) {
    Serial.print("Testing module nr "); Serial.println(i);
    Wire.beginTransmission(AnSet[i]);
    Wire.write(0x04);
    Wire.endTransmission();
    Wire.requestFrom(AnSet[i], 5);
    rv0=Wire.read();
    rv0=Wire.read();
    rv1=Wire.read();
    rv2=Wire.read();
    rv3=Wire.read();
 
    volt[0] = rv0 * 0.11171;
    volt[1] = rv1 * 0.0556;
    volt[2] = rv2 * 0.02272;
    volt[3] = rv3 * 0.02025;

[...]

PR #1849, opened 2022-04-29
add a global define that can disable Wire.setClock()

Bodobolero:
This pull request proposes to introduce a new global
`#define U8x8_DO_NOT_SET_WIRE_CLOCK`
which disables the Wire.setClock() invocations in     [  Wire.setClock(u8x8->bus_clock); ](https://github.com/olikraus/u8g2/blob/master/cppsrc/U8x8lib.cpp#L1345) and https://github.com/olikraus/u8g2/blob/7014e1d9a69883214b9bbcd64ccf0f7fd8af7964/cppsrc/U8x8lib.cpp#L1378.

Rationale:
while resolving problems over the last few weeks I have stumbled upon multiple people who could only resolve the problem by patching the u8g2 library and removing the setClock() invocations.

The scenario is: the I2c bus used for the monochrome display is also used for other devices.
The libraries of these other devices are initialised and used before the monochrome display is initiating the wire communication.

References:

1) 
https://github.com/arduino-libraries/Arduino_SensorKit/issues/21#issuecomment-1104030135

2) 
https://github.com/olikraus/u8g2/issues/1413#issue-816917931

3)
https://github.com/olikraus/u8g2/issues/1413#issuecomment-791354453

4)
https://github.com/olikraus/u8g2/issues/1134

5)
https://github.com/olikraus/u8g2/discussions/1749

P.S. the suggested solution to call setBusClock() before begin() did not work for me.
My setup:
I use an Arduino Nano RP2040 Connect.
The board already has crypto, wifi and accelerometer connected to the single I2C bus and uses firmware to control the wifi module.
Wifi stopped working as soon as I used u8g2 with the board and an Oled transaction set the clock speed.
I tried multiple different clock speeds with setBusClock() (400k, 100k, 1000k, ...) without success.

### DIFF
--- a/cppsrc/U8x8lib.cpp
+++ b/cppsrc/U8x8lib.cpp
@@ -1341,8 +1341,11 @@ extern "C" uint8_t u8x8_byte_arduino_hw_i2c(U8X8_UNUSED u8x8_t *u8x8, U8X8_UNUSE
     case U8X8_MSG_BYTE_START_TRANSFER:
 #if ARDUINO >= 10600
       /* not sure when the setClock function was introduced, but it is there since 1.6.0 */
-      /* if there is any error with Wire.setClock() just remove this function call */
-      Wire.setClock(u8x8->bus_clock); 
+      /* if there is any error with Wire.setClock() just remove this function call by */
+      /* defining U8x8_DO_NOT_SET_WIRE_CLOCK */
+#ifndef U8x8_DO_NOT_SET_WIRE_CLOCK
+      Wire.setClock(u8x8->bus_clock);
+#endif 
 #endif
       Wire.beginTransmission(u8x8_GetI2CAddress(u8x8)>>1);
       break;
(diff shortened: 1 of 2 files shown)


olikraus:
looks ok to me

olikraus:
Note: I have made U8x8 fully uppercase (U8X8...)

dejavu1987:
I am using a MCU6050 gyro censor that also uses I2C along with an SSD1306 OLED using U8G2.
I get this error when trying to read from gyro
`[Wire.cpp:499] requestFrom(): i2cWriteReadNonStop returned Error 263`
I have tried adding the flag introduced here, but did not change anything.
As usual, I have `Wire.begin()`  call in my setup function 
I am using ESP32-S3 
I2C_CLK 2
I2C_SDA 3

When I remove U8G2 the read from gyro does not have any errors.

olikraus:
Not sure how this comment is related to this PR.
Consider to create an issue with a precise error description. Also note, that Wire.cpp is not a u8g2 function, so maybe you should ask ESP32 people. The fact that the removal of u8g2 solve the problem, does not mean that u8g2 is the root cause for the same. U8g2 uses Wire.cpp and works reliable with it. If u8g2 missused Wire.cpp in any way, then i am happy to fix this.

PR #1854, opened 2022-05-03
Cmake: Test for register_component, if not found use std cmake add library

rivimey:
I wanted to use u8g2 in an environment in which register_component didn't exist, so came up with this simple patch. Hope it's useful. I'm not a cmake expert: if there's a better way I'm all ears! :-)


olikraus:
I am also not a cmake export... just pressing the merge button...

PR #1855, opened 2022-05-03
Check and extend content of ssd132x.txt to new file ssd13xx.md covering more controllers

rivimey:
I've been looking at supporting an SSD1320 based display with 320x132 4-bit greyscale and got into updating the document ssd132x.txt, checking the details and making some small fixes (e.g. some controllers have two 'disp off' rather than 'disp off' / 'disp on' and renaming  a few cells where that made more sense to me (e.g. rather than 'mode' I have 'norm disp' and 'inverse disp'

I converted it to markdown because it seems a safer and easier form to work in, and extended it with details for the 130x and 131x controllers. Consequently I made a new file, ssd13xx.md. I haven't deleted the original, though fullness of time that would be sensible.

I found reordering the table columns in Excel somewhat easier, so there is currently an .xlsx file for this; do you want it?

I have also assembled a bunch of the original datasheets from the public web; I presume you don't want this or they would be there already.


olikraus:
Wow, amazing work and greate addon for the existing table. 
Did you encounter any consequences out of this table? Just wondering why you created the same? Typical u8g2 users do not care much about the specific display commands required in the background...

rivimey:
I did note a few errors in the original table (I think I mentioned most in my initial post) but nothing that wasn't very obvious once you looked in detail. Mostly, I just got interested in understanding how it worked -- blame aspergers!

I was also trying to understand how a controller that claimed support for 160x160 could support my 320x132 display as that didn't make sende. I now know there are two controllers!

You didn't answer whether the xlsx file, or the set of pdf datasheets, was useful so I presume not.

olikraus:
> You didn't answer whether the xlsx file, or the set of pdf datasheets, was useful so I presume not.

I think it is good as it is :-)

PR #1856, opened 2022-05-03
First untested version of Surenoo SSD1320 based 320x132x4 display for RP2040 and/or Zephyr (review)

rivimey:
Display P/N: SOG320132A_M383

Surenoo Display datasheet: http://surenoo.tech/download/03_SOL/0302_SOG/SOG320132A_M383.pdf
  (a bit brief, but adequate)
Surenoo SSD1320 datasheet: http://surenoo.tech/download/03_SOL/0302_SOG/Controller/SSD1320.pdf
  (includes command list)

Two interfaces present: one for the Raspberry Pi RP2040 SDK, using e.g. gpio_set_function(),
and one which I intended to be generic for Zephyr-RTOS, but which my lack of understanding
is hampering. I think the pico-sdk versions are essentially ok, but the zephyr ones are not
as yet.

I have copied over the group of functions u8g2_m_40_17_2 et al and I think adjusted them
properly, but I do not understand the choice of '17' here. Perhaps this area needs checking?
I have also put #if 0 / #endif the *nhd_powersave1_seq arrays because they weren't used. I
have no idea if they are correct for the 1320.

### DIFF
--- a/csrc/u8x8_d_ssd1320_320x132.c
+++ b/csrc/u8x8_d_ssd1320_320x132.c
@@ -1,0 +1,650 @@
+/*
+
+  u8x8_d_ssd1320.c
+  
+  Universal 8bit Graphics Library (https://github.com/olikraus/u8g2/)
+
+  Copyright (c) 2020, olikraus@gmail.com
+  All rights reserved.
+
+  Redistribution and use in source and binary forms, with or without modification, 
+  are permitted provided that the following conditions are met:
+
+  * Redistributions of source code must retain the above copyright notice, this list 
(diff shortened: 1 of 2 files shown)


olikraus:
I need to refuse. Most of the will be (u8x8_m functions) code is generated and the csrc folder should not contain target specific functions (gpio and delay). 
I suggest to create an issue instead. I will add your code and run the code generator. After this you could do the fine tuning. Would this be ok?

rivimey:
Created issue https://github.com/olikraus/u8g2/issues/1858, thanks.

PR #1866, opened 2022-05-17
add ssd1681 driver to lib, removed constants in codebuild, changed er…

pikot:
…ror to printf/exit sequence (it should works on mac and linux at the same time, mac doesn't have error.h file),


olikraus:
There are some nice updates included, but all in all I decided not to include the pr because of the busy flag. Users of u8g2 already complain about the code size and the busy flag will again increase the same for all users.

PR #1929, opened 2022-08-06
 mingw sys/sdl : "undefined reference to `u8x8_GetMenuEvent'"

sudobob:
When building apps in  sys/sdl under MinGW,  gcc was failing with "undefined reference to `u8x8_GetMenuEvent'" 
Using MinGW gcc version 6.3.0 (MinGW.org GCC-6.3.0-1)


olikraus:
Thanks,  I need to check this.

olikraus:
I was in need to revert this, because it breaks u8g2. We need to look for a different solution.

PR #2009, opened 2022-11-13
Checking the definition of MUI_MAX_TEXT_LENGTH

kasitoru:
This will allow to change the value of MUI_MAX_TEXT_LENGTH at the compilation stage without editing the source code.


olikraus:
ok...

PR #2114, opened 2023-02-24
handle '\r'

bartlomiejcieszkowski:
carriage return wasn't handled
spinner is now correctly working

tried making spinner, instead of returning to back of line, it kept writing on and on till going off the screen

minimal loop for reproducing:
```
void loop() {
  u8x8.setFont(u8x8_font_chroma48medium8_r);
  u8x8.setCursor(0, 0);
  int length = 60;
  char spinner[] = "|/-\\";
  for (int i = 0; i < length; i++) {
    // do something
    u8x8.printf("%c\r", spinner[i%4]);
    delay(200);
  }
}
```

### DIFF
--- a/cppsrc/U8x8lib.cpp
+++ b/cppsrc/U8x8lib.cpp
@@ -72,6 +72,10 @@ size_t U8X8::write(uint8_t v)
     ty+=dy;
     tx=0;
   }
+  else if ( v == '\r' )
+  {
+    tx=0;
+  }
   else
   {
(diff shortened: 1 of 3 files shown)


olikraus:
I don't see a reason to merge this..

CodeAsm:
this worked for me (altho there are some bits in the logging i miss, maybe ill spin my own version)
```c
void wait() {
  int length = 32;
  char spinner[] = "|/-\\";
  for (int i = 0; i < length; i++) {
    // do something
    u8x8log.print(spinner[i%4]);
    u8x8log.print('\r');
    delay(100);
  }
}
```

Issue #377, opened 2017-10-04
Explicitly Define HW SPI Interface Pins - MOSI & SLCK


neilpanchal:
Hello,

`U8G2_SSD1306_128X64_NONAME_F_4W_HW_SPI display1(U8G2_R0, /* cs=*/ 5, /* dc=*/ 2, /* reset=*/ 4);`

Doesn't allow the user to define the MOSI and SCLK pins in the constructor. 

It is especially useful on boards with 2 or more SPI modules such as on the ESP32. I probed around with a scope and found out that the constructor selects VSPI as opposed to HSPI module. 

A couple of questions to try to understand the rationality behind protecting the user from defining the pins themselves.

1. Is it possible to connect displays on both HSPI and VSPI lines? I understand that I can use the chip-select (CS) pin to select between displays, however, I am trying to avoid robbing clock cycles and therefore reducing the framerate (not sure how true my claim is...but I just want to try it out). 

2. I am curious why there are some drivers whose constructors allow for data and clock pin definition and others whose constructors don't. I am using SSD1306.

3. Is this something that needs to be handled by the ESP32 HAL? I haven't had a chance to delve into the U8G2 code, or the ESP32 IDF and/or espressif32 for Arduino.

Thank you!

olikraus:
First of all, let me note that ESP32 is not officially supported (or at least is not tested)

> A couple of questions to try to understand the rationality behind protecting the user from defining the pins themselves.

There is no protection, it is simply impossible for most of the uC to change the SPI/I2C lines. Notable exceptions include the ESP systems. 
Even more difficult: Arduino SPI and Wire procedures usually do not require pin numbers (and also will not accept/ignore them for most of the Arduino boards).

For I2C (but not for SPI), there is already an u8g2 extension:
```
U8G2_SSD1305_128X32_NONAME_F_HW_I2C(rotation, [reset [, clock, data]])
```

It is possible to add clock and data lines as optional arguments to the end of constructor.
Lets take the Wire lib as an example.

The official page does not at all allow you to use any other pins for I2C:
https://www.arduino.cc/en/Reference/WireBegin

However, there is an ESP8266 specific extension which is used by u8g2 in the special case of ESP8266
https://github.com/esp8266/Arduino/blob/master/libraries/Wire/Wire.h#L53

For SPI, no such extension is known to me and also not implemented.

[...]

PR #2123, opened 2023-03-01
Bugfix ESP32: I2C pin defs used instead of the SPI pin defs (fixes pin remapping for HW SPI on ESP32 platform)

kazetsukaimiko:
Follow-up from:
https://github.com/olikraus/u8g2/issues/377#issuecomment-1448419338

### Platform
Arduino / ESP32-S3 
### Display
OLED SSD1309 (SPI)

### Summary

A custom PCB requires non-standard SPI pin assignments for an SSD1309 OLED display driven by an ESP32-S3.
Everything has been tested functional with bit-banged SW_SPI (U8G2_SSD1309_128X64_NONAME2_F_4W_SW_SPI constructor) on the required "custom" SPI pin assignments.  However, the display manages a very poor 18-19fps frame rate:

software_bb_20%

It appears in the U8G2 code that there is some proviso for manual pin assignments on the HW_SPI functionality.  However, after creating a custom constructor to utilize this functionality, some notable bugs were discovered in the U8G2 HW_SPI setup functions: namely, that the "manual pin assigments" query the I2C pin assignment registers, instead of the SPI pin assignment registers (which only the latter are set by the custom HW_SPI constructor).
This error results in the manual SPI pin assignments being effectively inaccessible from the U8G2 constructors.

After correcting the "I2C pin assignments" to "SPI pin assignments", the U8G2 library can utilize the desired manual pin assignments with the SSD1309 OLED screen, showing a huge marked improvement in framerate to approximately 230fps (more than 10 times faster than SW_SPI), as seen here:

230fps_20%

One further improvement that can be made to HW_SPI is to utilize a relatively unknown SPI function for bulk SPI writes on the ESP32, namely "SPI.writeBytes".  (It has been mainstream in Arduino-ESP32 code since at least 2017.)  Comments in the U8G2 code indicate that the "SPI.transfer" function has been found to overwrite the input buffer; "SPI.writeBytes" does not overwrite the input buffer (this has been tested).
Switching to "SPI.writeBytes" results in a further improvement in frame rate to 374fps, an easy 35% improvement from the "byte transfer" frame rate, as seen here:

324fps_20%

### DIFF
--- a/cppsrc/U8x8lib.cpp
+++ b/cppsrc/U8x8lib.cpp
@@ -885,6 +885,12 @@ extern "C" uint8_t u8x8_byte_arduino_hw_spi(u8x8_t *u8x8, uint8_t msg, uint8_t a
   {
     case U8X8_MSG_BYTE_SEND:
       
+
+
+#if defined(ESP_PLATFORM)
+      //T.M.L 2023-02-28: use the block transfer function on ESP, which does not overwrite the buffer.
+      SPI.writeBytes((uint8_t*)arg_ptr, arg_int);  
+#else    
       // 1.6.5 offers a block transfer, but the problem is, that the
       // buffer is overwritten with the incoming data
       // so it can not be used...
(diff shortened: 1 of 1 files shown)


kazetsukaimiko:
Credit / thanks to @WebDust21 for doing the deep dive on this.

PR #2162, opened 2023-04-20
Update CMakeLists.txt

Eonassis:
missing no link file CMakelists.txt to run ESP32 C3 display,  It is an honor to contribute to this beautiful project.


olikraus:
Thanks, indeed i do depend on the community to update this file. :-)

PR #2166, opened 2023-04-27
Linux Framebuffer support

QB4-dev:
Hey
I've added direct linux framebuffer support. Linux kernel supports many monochrome displays by kernel drivers like ssd1306fb, and they are mapped in system as standard framebuffer devices like /dev/fb0. It is possible to connect there system console or draw by writing directly to device. u8g2 port may be very useful if you need to display simple UI on Linux framebuffer; 
I can't belive that no one did it before


olikraus:
looks good so far, I guess I need to check this once

PR #2201, opened 2023-06-10
bar with customized width

njuFerret:
When using a bar graph as an indicator, such as for PWM's duty ratio, it can be challenging to draw the graph at a specific position and width. Modifications has been made in the source code to implement such a scenario.

### DIFF
--- a/csrc/mui_u8g2.c
+++ b/csrc/mui_u8g2.c
@@ -1106,10 +1106,10 @@ uint8_t mui_u8g2_u8_bar_wm_mse_pf(mui_t *ui, uint8_t msg)
     case MUIF_MSG_CURSOR_SELECT:
     case MUIF_MSG_VALUE_INCREMENT:
       (*value)+=step;
-      if ( *value > max ) *value = min;
+      if ( *value > max ) *value = max;
       break;
     case MUIF_MSG_VALUE_DECREMENT:
-      if ( *value >= min+step ) (*value)-=step; else *value = max;
+      if ( *value >= min+step ) (*value)-=step; else *value = min;
       break;
     case MUIF_MSG_CURSOR_LEAVE:
       break;
--- a/csrc/mui_u8g2.c
+++ b/csrc/mui_u8g2.c
@@ -913,7 +913,7 @@ uint8_t mui_u8g2_u8_min_max_wm_mud_pf(mui_t *ui, uint8_t msg)
       {
         (*value)++;
         if ( *value > max )
-          *value = min;
+          *value = max;
         return 1;
       }
       break;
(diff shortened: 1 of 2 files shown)


njuFerret:
Forgot to add a macro to keep compatibility
`#define MUIF_U8G2_U8_MIN_MAX_STEP(id, valptr, min, max, step, flags, muif) MUIF_U8G2_U8_MIN_MAX_STEP_WIDTH(id, valptr, min, max, step, max, flags, muif)`

njuFerret:
the macro patch, as well as  value with boundary in `mui_u8g2_u8_bar_wm_mse_pf` function.
[bar_macro_fix.patch](https://github.com/olikraus/u8g2/files/11713712/bar_macro_fix.patch)

olikraus:
So the "width" argument is the total width of the bar graph.
I also think that the scale argument is kind of obsolete, because we could just multiply width with 2 or 4 instead.
All in all I more think about creating a second bar graph instead of updating the existing one.

olikraus:
One more doubt: You have disabled the wrap around to the min value. This would be fine for the "mud" edit mode, but would be an issue with the "mse" edit mode. Instead we should have an option here instead of hard coding this. I think your PR addresses multiple problems, which should be fixed independently.

njuFerret:
> So the "width" argument is the total width of the bar graph. I also think that the scale argument is kind of obsolete, because we could just multiply width with 2 or 4 instead. All in all I more think about creating a second bar graph instead of updating the existing one.

in fact, the "width" argument is the type of `uint8_t`, when small screen,like 12864, the scale argument maybe no use. but for some larger display, this argument maybe useful.

njuFerret:
> One more doubt: You have disabled the wrap around to the min value. This would be fine for the "mud" edit mode, but would be an issue with the "mse" edit mode. Instead we should have an option here instead of hard coding this. I think your PR addresses multiple problems, which should be fixed independently.

a rotary encoder was employed for user input, where `mse` mode isn't the candidate. therefore, the modification was not been tested. I think it's a good idea provide user some option to reach 

> So the "width" argument is the total width of the bar graph. I also think that the scale argument is kind of obsolete, because we could just multiply width with 2 or 4 instead. All in all I more think about creating a second bar graph instead of updating the existing one.

ye,  "max" used as total width of bar graph,  was replaced by "width" argument. The "scale" arg maybe useful for large display, due to "width" defined as "uint8_t".

olikraus:
I have recreated a new field function instead: https://github.com/olikraus/u8g2/wiki/muiref#fixed-width-bar-graph-uint8

So closing this PR without merge.

njuFerret:
> I have recreated a new field function instead: https://github.com/olikraus/u8g2/wiki/muiref#fixed-width-bar-graph-uint8
> 
> So closing this PR without merge.

awesome!

PR #2217, opened 2023-07-06
Support for ST7567 LW12832

shezik:
Hi olikraus,

This display has a few interesting traits. It's generally consistent with ST7567 JLX12864, but its physical resolution is padded from 128x32 to 142x36 with weird behaving margins. Since the bits representing the margins are outside display area, there's no way to overwrite them, and they become random garbage data whenever the controller (consequently the DDRAM) is power cycled.

The newly added `static void st7567_lw12832_clear_ddram(u8x8_t *u8x8)` function is to address this issue by setting DDRAM to all zero while ignoring `u8x8->x_offset`, thus clearing everything including the margin bits.

Code is tested using `U8G2_ST7567_LW12832_F_4W_SW_SPI` and it works fine. Changelog and examples haven't been updated yet.


olikraus:
Thanks for your contribution, I will merge the pr.

PR #2222, opened 2023-07-10
Add support for GP1294AI

izilzty:


ccoenen:
should this also be listed under controllers in the README?

olikraus:
Yes it is: https://github.com/olikraus/u8g2/wiki/u8g2setupcpp#gp1294ai-256x48

PR #2277, opened 2023-10-05
Update u8x8_d_framebuffer.c

QB4-dev:
screensize variable was not set


olikraus:
Nice finding... Well, this code was contributed, so I never tested this...

QB4-dev:
That's my code, that's why i found it :P

Thank you for merge

olikraus:
:-D 

ah ok, then: Many thanks for the update!

PR #2445, opened 2024-05-17
Use correct resolution register value. Fixes dim artifacts on edge of 128x160 screens

Muny:


olikraus:
looks useful, hope it will not break anything. Thanks for your PR.

PR #2456, opened 2024-06-13
add hc32 driver u8g2 example

KemGong:
add mcu using hardware spi driver u8g2 shows example


olikraus:
Thanks for providing the example.
Is there any link to the description of the HC32 controller?

KemGong:
Of course, here is the link https://www.xhsc.com.cn/Productlist/info.aspx?itemid=1852

PR #2472, opened 2024-07-11
Fixing support for registering this library as an ESP-IDF component

fherrera124:


olikraus:
THanks, actually i am not using that file, so I hope that this will help as intended.

Issue #2468, opened 2024-07-05
MUI: API Functions for a 'back' button


microbmen:
Hello @olikraus ,  

I have implemented up and down buttons,  prevField and nextField respectively.  
I have implemented a select button,  sendSelect, 

What I cant figure out is how to implement a back button. 

Requirement:
API call which does a sendBack() function, returning to the previous MUI_FORM on the previous cursor position.

We can use gotoForm as a back button, and track the form ID needed depending on the current form displayed, however gotoForm requires an initial position, which gives odd UI/UX performance.  
An API call for currentCursor(), currentForm(), would also suffice, as then we could retain the position and form independent and use gotoForm anyway we'd like.  

Unless I am missing something, this seems to be a needed feature for 4 button control (up/down , previous/next) 

Thanks!

qianfan-Zhao:
I am learing 'mui' example this days. All the menu entry has a ui return button such as this example:

```c
/* number entry demo */
MUI_FORM(10)
MUI_STYLE(1)
MUI_LABEL(5,10, "Number Menu")
MUI_XY("HR", 0,13)
MUI_STYLE(0)

MUI_LABEL(5,27, "Number [mse]:")
MUI_XY("IN",76, 27)
MUI_LABEL(5,41, "Number [mud]:")
MUI_XY("IH",76, 41)

MUI_XYAT("G1",64, 59, 0, " OK ")
```

The last line `MUI_XYAT` has a pointer value set which form will return to when this button is pressed, seems all the menu entries use this button to return to it's parent, I can not find an example which use hardware 'backup' button handle it.

I am interesting how to add a hardware backup button, maybe we can take more discussion.

qianfan-Zhao:
@microbmen Hi, I pushed a new PR #2479 try add back button, could you please try it?

microbmen:
ok, will try it.  i created a work around by setting a 'current menu' flag using MUI_AUX and then using gotoform.

olikraus:
Currently I do hesitate to include the above mentioned PR, because there is already a little bit of code, which should support this. I also think, that the cursor topic is not addressed by pr #2479 .
Let me look into this a little bit deeper.

olikraus:
Let me give a small background on this topic. There are two concepts in u8g2:
[...]

PR #2479, opened 2024-07-19
Add mui_Return to switch to parent window

qianfan-Zhao:


olikraus:
I don't see the need for this. It could be implemented in the app event loop itself by checking for the key and forcefully jumping to the parent page.

qianfan-Zhao:
Hi @olikraus, new version had pushed, please review again.

olikraus:
https://github.com/olikraus/u8g2/issues/2468

olikraus:
Currently I am working on a more complete solution for this, so I will not merge the PR.
The new solution will introduce a new back MUIF, which will allow to jump back to a previous form.

PR #2486, opened 2024-08-13
Support using the bitmap device in the Arduino library

matthijskooijman:
The header files in `csrc` and `cppsrc` contained various declarations related to the bitmap device backend (e..g the `u8g2_SetupBitmap` function and the `U8G2_BITMAP` class), but the definitions of those were in `sys/bitmap/u8x8_d_bitmap.c` and thus not included in default builds and in particular not the Arduino library.

This moves this code into `csrc`, leaving behind the TGA export code that relies on stdio.h, to make this bitmap backend usable in the Arduino library. It also adds an example for how to use this code in the Arduino library.

I tested:
 - The new example From within u8g2 with `Makefile.184.uno`, an unpacked Arduino 1.8.4 tarball and an offical Uno R3.
 - The same with `arduino-cli`, the `arduino:avr` core version 1.8.5 after generating an Arduino library with `tools/release/arduino/create_release.sh`
 - The examples in `sys/bitmap`

The Arduino example sketch outputs the following:

image

### DIFF
--- a/csrc/u8x8_d_bitmap.c
+++ b/csrc/u8x8_d_bitmap.c
@@ -9,7 +9,6 @@
 #include "stdlib.h"	/* malloc */
 #include "stdint.h"	/* uint16_t */
 #include "string.h"	/* memcpy */
-#include "stdio.h"	/* FILE */
 #include "u8g2.h"		/* because of u8g2_Setup... */
 
 /*========================================================*/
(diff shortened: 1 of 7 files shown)


olikraus:
hmmm isn't this very similar to https://github.com/olikraus/u8g2/wiki/u8g2reference#writebufferxbm

matthijskooijman:
> nice idea, so you just output the bitmap as a x/blank ASCII art, correct?

Yeah, that's just a way to actually do something with the pixel data, but the point of the example is to show how to access the raw pixels.

> hmmm isn't this very similar to https://github.com/olikraus/u8g2/wiki/u8g2reference#writebufferxbm

In a way, but that outputs XBM format, without accessing the pixels programmatically.

However, reading onwards, I see: https://github.com/olikraus/u8g2/wiki/u8g2reference#direct-access-buffer-api

Which is an API to directly access the rendering buffer tiles/pixels, which would actually also work perfectly for my usecase.

Thinking on this a bit more, I actually wonder what the point of the "bitmap" device is - u8g2 already keeps a buffer containing all pixels, which is then copied to another buffer inside the "bitmap" device, but there does not really seem to be any point (and it needs twice the memory).

I would say this could just as well use a "dummy" device (that just throws away pixels) and then use the above buffer API to access the upper buffer directly - for this example, but also for the non-arduino examples that use the bitmap device currently for TGA output (there could also be a simple TGA output function that reads from the buffer like writeBufferXBM does).

olikraus:
hmmm... you have removed the TGA write procedures, but those are needed for the automatic image generation.
Even if we keep the old file, the PR will break the tga image generation. I think more effort is required to fix this.

matthijskooijman:
> you have removed the TGA write procedures

They are not removed, but split off from the bitmap backend. They are still here: https://github.com/matthijskooijman/u8g2/blob/509954977b2d40c34dfc6f7a9a696910b93d88cd/sys/bitmap/common/u8x8_d_bitmap_tga.c


This does mean that TGA generation is not available in the Arduino version, but TGA is probably not relevant there anyway (and this particular implementation relies on stdio.h, which is not available in all architectures IIRC.

>  Even if we keep the old file, the PR will break the tga image generation. I think more effort is required to fix this.

The [hello_world](https://github.com/matthijskooijman/u8g2/tree/509954977b2d40c34dfc6f7a9a696910b93d88cd/sys/bitmap/hello_world) and [cimg_annotate_screenshot](https://github.com/matthijskooijman/u8g2/tree/509954977b2d40c34dfc6f7a9a696910b93d88cd/sys/bitmap/cimg_annotate_screenshot) examples still work for me. Is there something else that you think will break?

olikraus:
I will move this topic...

PR #2525, opened 2024-10-18
Fix implicit typecast from void* to uint8_t*

JohnZ03:
This implicit typecast cause error on some compilers, manually casting the void pointer before referencing fixes this issue.

image


olikraus:
Thanks!

Issue #2658, opened 2025-06-12
ESP-IDF build broken


mars3142:
Since the merge of #2655 (commit 18dc4a8294d7aa14ae58dd91b09ac332b68f4aa5) the code can't be compiled with ESP-IDF.

The error messages are: 

    u8g2/csrc/u8x8_u8toa.c:96:15: warning: variable 'buf' set but not used [-Wunused-but-set-variable]
    u8g2/csrc/u8x8_u8toa.c:108:1: error: control reaches end of non-void function [-Werror=return-type]

This method ```const char *u8x8_u8tox(uint8_t v, uint8_t d)``` doesn't return any value and ```buf[3]``` is unused.

olikraus:
yes, indeed

mars3142:
Sorry, next error:

    /csrc/u8g2_font.c:1565:1: error: "/*" within comment [-Werror=comment]

olikraus:
thanks & fixed

PR #2659, opened 2025-06-12
mui: uint8_t hex min/max bug #2658

koendv:
bugfix


mars3142:
It's related to #2658.

PR #2712, opened 2025-10-16
add st7305_yuying_200x200 and st7305_yuying_168x384 driver

lingyunya:
Two reflective screens driven by **st7305** have been added. They are the **200x200** resolution screen from osptek and the **168x384** resolution screen from osptek.


olikraus:
What is the purpose/benefit of you PR? 
The constructor is already present without the yuying prefix.
The second problem is the missing update of codebuild.c

lingyunya:
> What is the purpose/benefit of you PR? 
> The constructor is already present without the yuying prefix.
> The second problem is the missing update of codebuild.c 
> 
Regarding the submitted information, there was no corresponding code in the repository when I was writing the driver code. By modifying the existing driver and debugging, I found that the screen from the manufacturer I am using cannot be driven properly, so I wrote drivers for two corresponding screens. I will complete the modification of codebuild. c as soon as possible, thank you.

PR #2733, opened 2025-12-11
Add conditional definition for u8g_sdl_multiple

marshfolx:
Add compile definition U8G2_U8G_SDL_MULTIPLE to adjust SDL window size and pixel size.

The default value is:

```c
int u8g_sdl_multiple = 3;
```

When the value is set to 6, the window size should be 2x bigger.

image


marshfolx:
@olikraus any problem with this PR?

olikraus:
thanks for the PR... no problem...
I just don't have much time for u8g2...

PR #2753, opened 2026-02-15
fix: several compilation issues under LLVM and more

basilfx:
I am [upgrading](https://github.com/RIOT-OS/RIOT/pull/22081) the U8g2 package version in RIOT-OS. This PR fixes some of the compilation errors I encountered.

I have been able to test this against some displays and board I own. But the warnings below occur in drivers and functionality I cannot test.

## Comparison of integers of mixed signs:

This fails for MSP-430 (16-bit targets). Casting `rad` to `u8g2_int_t` solves it.

```
u8g2_arc.c: In function 'u8g2_draw_arc':
u8g2_arc.c:78:16: error: comparison of integer expressions of different signedness: 'u8g2_int_t' {aka 'short int'} and 'unsigned int' [-Werror=sign-compare]
   78 |     else if (d < 2 * (rad - y))
      |                ^
```

## Unused variables (LLVM):

These variables are assigned to, but never read from.

```
u8x8_d_st7305.c:279:12: error: variable 'x' set but not used [-Werror,-Wunused-but-set-variable]
  uint16_t x;
           ^
u8x8_d_st7305.c:395:12: error: variable 'x' set but not used [-Werror,-Wunused-but-set-variable]
  uint16_t x;
           ^
u8x8_d_st7305.c:609:12: error: variable 'x' set but not used [-Werror,-Wunused-but-set-variable]
  uint16_t x;
           ^
u8x8_d_st7302.c:418:12: error: variable 'x' set but not used [-Werror,-Wunused-but-set-variable]
  uint16_t x;
           ^
```

## Memory alignment requirement (LLVM):

Using `memcpy` instead resolves this issue.

```
u8x8_d_sdl_128x64.c:53:13: error: cast from 'uint8_t *' (aka 'unsigned char *') to 'uint32_t *' (aka 'unsigned int *') increases required alignment from 1 to 4 [-Werror,-Wcast-align]
   53 |       ptr = (uint32_t *)(((uint8_t *)(u8g_sdl_screen->pixels)) + offset);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

## Mark variable as unused (LLVM):

`max_ptr` is only used in `assert(..)`, which may be a no-op if assertions are disabled. Mark it as unused resolves this.

```
u8g2_ll_hvline.c:79:12: error: unused variable 'max_ptr' [-Werror,-Wunused-variable]
   79 |   uint8_t *max_ptr = u8g2->tile_buf_ptr + u8g2_GetU8x8(u8g2)->display_info->tile_width*u8g2->tile_buf_height*8;
      |            ^~~~~~~
1 error
```

### DIFF
--- a/csrc/u8g2_ll_hvline.c
+++ b/csrc/u8g2_ll_hvline.c
@@ -77,6 +77,7 @@ void u8g2_ll_hvline_vertical_top_lsb(u8g2_t *u8g2, u8g2_uint_t x, u8g2_uint_t y,
   uint8_t or_mask, xor_mask;
 #ifdef __unix
   uint8_t *max_ptr = u8g2->tile_buf_ptr + u8g2_GetU8x8(u8g2)->display_info->tile_width*u8g2->tile_buf_height*8;
+  (void)max_ptr;	/* silence unused variable warning if asserts are disabled */
 #endif
 
   //assert(x >= u8g2->buf_x0);


olikraus:
thanks for the work so far.
mostly valid points, however the unused x variable seems to be a bug in the low level driver. Removal is not the correct solution.
I would accept the PR, but I need to review the unused x variable more in detail.

Would it be possible to split the PR? I mean a PR without the fix for the unused variables? Or at least if the code for the unused x vars would be in comments. That would be fine also.

Issue #2756, opened 2026-03-01
code cleanup: PR 2753


olikraus:
https://github.com/olikraus/u8g2/pull/2753

olikraus:
ok... could should be clean now....
@basilfx   thanks for the hint!

basilfx:
Thanks @olikraus for looking into this.

I have re-checked, but there still one left: [this one](https://github.com/basilfx/u8g2/commit/2da3a2f22c45ba22a2b05f9d23a85df5a5dfa4c7#diff-f028efabdb29afede037e6c8cbb5b61a42d04c942cdd8ac2fbd1346d01a540e8) (in csrc/u8x8_d_s1d15e06.c).

Reading the [datasheet](https://datasheet4u.com/pdf-down/S/1/D/S1D15E06_Epson.pdf) (section 6.2.3) makes me believe this `y` values is also redundant, as the direction is automatically incremented after read/write ("Whichever direction is chosen, increment is carried out by positive one (+1) after write or read operation.").

Again, totally not sure, just want to provide a hint if that helps ;-)

(I'll check the other fixes tonight: the build system of [RIOT-OS](https://ci.riot-os.org/details/8cb94f383c4248cdb60e2ef6b19be443) stops after the first error, so I have to add a 'temporary' fix to see if that succeeds.)

olikraus:
> I have re-checked, but there still one left: [this one](https://github.com/basilfx/u8g2/commit/2da3a2f22c45ba22a2b05f9d23a85df5a5dfa4c7#diff-f028efabdb29afede037e6c8cbb5b61a42d04c942cdd8ac2fbd1346d01a540e8) (in csrc/u8x8_d_s1d15e06.c).

I think you can just comment out the y calculation

PR #2761, opened 2026-03-09
fix: comment unused y in s1d15e06

basilfx:
Per #2756, I commented the unused `y` variable in s1d15e06.

This seems to address the last compilation issue for [updating](https://github.com/RIOT-OS/RIOT/pull/22081) U8g2 in RIOT-OS.

I don't have this display, so I am unable to test this.


olikraus:
ah, already long time back, but if I remember, this should be ok.
Thanks for the PR.

Issue #2135, opened 2023-03-19
Please add support ELW0901AA (LD7032)


iggymayer:
https://mega.nz/file/AF0QSAYR#xLxlIQQKIUClCdAZPUkWpPQ64zNi-doQkAz_ZGgTioM
Please add support ELW0901AA,Driver IC:LD7032.
If you can,please support both I2C & SPI interfaces,Thank you!

olikraus:
The init sequence in the above datasheet implies, that you will drive the display logic power with 1.8V and you will provide an external 18V power supply for the OLED itself, correct?

olikraus:
I have created a new beta release with the new constructor "U8G2_LD7032_128X36..."

Can you check whether the new constructor works?



You can download the latest U8g2 beta release from here: https://github.com/olikraus/U8g2_Arduino/archive/master.zip
Arduino IDE:
1. Remove the existing U8g2_Arduino library (https://stackoverflow.com/questions/16752806/how-do-i-remove-a-library-from-the-arduino-environment)
2. Install the U8g2_Arduino Zip file via Arduino IDE, add zip library menu (https://www.arduino.cc/en/Guide/Libraries).

PlatformIO:
`platformio.ini` (https://docs.platformio.org/en/latest/projectconf/section_env_library.html#lib-deps) should include 
```
lib_deps =
  u8g2=https://github.com/olikraus/U8g2_Arduino/archive/master.zip
```

iggymayer:
I will test it after receiving the PCB next month，Thank you!

olikraus:
Any feedback on this?

iggymayer:
Set Address 0x6E
Can be lit but cannot be displayed normally.

olikraus:
which constructor did you use? which code/example did you run?

iggymayer:
> which constructor did you use? which code/example did you run?

> Hello World
#include <Arduino.h>
#include <U8g2lib.h>
#include <Wire.h>

U8G2_LD7032_128X36_F_HW_I2C u8g2(U8G2_R0, 9, 8, 1);

void setup(void)
{
  u8g2.setI2CAddress(0x6E);
  u8g2.begin();
}

void loop(void)
{
  u8g2.clearBuffer();                   // clear the internal memory
  u8g2.setFont(u8g2_font_ncenB08_tr);   // choose a suitable font
  u8g2.drawStr(10, 10, "Hello World!"); // write something to the internal memory
[...]

PR #2785, opened 2026-06-10
fix ELW0901AA issues #2135

iggymayer:
fix issues #2135
I2C & SPI test passed
**Notice: If using I2C, the I2C buffer size needs to be increased: `Wire.setBufferSize(600);`**
IMG_2097


olikraus:
What is the issue with the buffer size? The problem is, that this is a restriction of the Arduino framework.
U8g2 architecture tries to deal with the limited buffer.

olikraus:
where exactly is the i2c buffer size increased?

iggymayer:
when using the hardware I2C interface, several bright dots appear on one side of the screen (not damaged). The data being sent exceeds the default size of the Arduino Wire library's send buffer. Due to my limited abilities, this is the only solution I can think of at the moment.

```
#include <Arduino.h>
#include <U8g2lib.h>

#ifdef U8X8_HAVE_HW_SPI
#include <SPI.h>
#endif
#ifdef U8X8_HAVE_HW_I2C
#include <Wire.h>
#endif

#define RES_PIN 21
#define HV_EN 1

U8G2_LD7032_128X36_F_HW_I2C u8g2(U8G2_R0, RES_PIN);

void setup(void) 
{
    pinMode(HV_EN, OUTPUT);
    digitalWrite(HV_EN, HIGH);

    Wire.setBufferSize(600);

    u8g2.begin();
    u8g2.setFontPosTop();
}

void loop(void) 
{
    u8g2.clearBuffer();
    u8g2.setFont(u8g2_font_spleen6x12_mf);
    u8g2.drawStr(0,0,"Hello World!");
    u8g2.drawBox(99, 0, 99, 36);
    u8g2.sendBuffer();

    delay(1000);
}

```

olikraus:
the  u8x8_cad_ld7032_i2c procedure will break this down to smaller blocks, except for the init sequence.
If the init sequence is more than 32 bytes it might be a problem.

https://github.com/olikraus/u8g2/blob/ab9e48b2228351e9476682a70b7f3ee4909cd585/csrc/u8x8_cad.c#L704

olikraus:
A solution could be to reduce the size of u8x8_d_ld7032_128x36_init_seq.
We could also have two arrays and send both arrays one after the other.

Maybe that is also the reason why the 60x32 never worked for me with I2C

I could also try to update the cad procedure: The buffer could be flushed with every new command. (buffer flush probably is not allowed during arg transmission). Let me create an issue for this.

I think fixing u8x8_cad_ld7032_i2c would be the most compatibe solution from my perspective (provided, that the init sequence is the problem)

olikraus:
Here is my suggestion: I will try to fix u8x8_cad.c, then we can include your LD7032 code.
issue: https://github.com/olikraus/u8g2/issues/2789

iggymayer:
Okay, thank you.

PR #2786, opened 2026-06-14
fix flipmode: OEL1M0033WE

iggymayer:
fix flipmode: OEL1M0033WE
test passed


olikraus:
hmm the diff view is useless in github. Not sure how i can review this. What is the exact change you did here?

iggymayer:
optimize serial interface timing.
fix character overlap in flip mode.

olikraus:
i think this looks good, thanks

PR #2788, opened 2026-06-25
Update Makefile

iggymayer:
when executing `make clean`, the files generated in the `codebuild` directory are cleared.


olikraus:
> when executing make clean, the files generated in the codebuild directory are cleared.

that is the purpose...

PR #2795, opened 2026-07-18
Read binary P4 PBM data in the font converter

94xhn:
## What changed

`tools/font/pbm/pbm.c` recognizes P4 files, but only the P1 branch reads raster pixels. P4 input therefore returns success with an uninitialized bitmap.

This change:

- opens PBM input in binary mode;
- reads P4 raster bytes in PBM's MSB-first order while preserving the tool's existing internal bitmap layout;
- rejects truncated P4 input and propagates the read failure from the CLI.

P1 parsing and output are unchanged.

## Tests

- MinGW GCC 8 and Linux GCC: `-std=c99 -Wall -Wextra -Werror`.
- Real-source 10x2 P1/P4 equivalence fixture, including `0x1a` and `0x0d` binary bytes: pixels match.
- Truncated P4 fixture: rejected; Linux AddressSanitizer/UBSan pass.
- Existing `Scroll-o-Sprites.pbm` P1 conversion: fixed and baseline output match across 5,416 lines.

### DIFF
--- a/tools/font/pbm/pbm.c
+++ b/tools/font/pbm/pbm.c
@@ -163,6 +163,24 @@ int pbm_ReadFP(pbm_t *pbm)
       }
     }
   }
+  else
+  {
+    uint16_t x, y;
+    int c = 0;
+    for( y = 0; y < pbm->h; y++ )
+    {
+      for( x = 0; x < pbm->w; x++ )
+      {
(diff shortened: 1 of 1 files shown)


olikraus:
ok, cool, I never thought anyone is interested in that small piece of code.

PR #2799, opened 2026-08-05
fix(bdfconv): prevent assertion failure when generating very large Unicode fonts (full CJK)

createskyblue:
## Problem

`bdfconv` aborts with an assertion when generating a font with a very large
number of Unicode glyphs (e.g. the full CJK block U+4E00–U+9FFF, ~21000 glyphs):

```
RLE Compress: Unicode lookup table len=210, written entries=209
Assertion failed!
File: bdf_rle.c, Line 898
Expression: unicode_lookup_table_len == unicode_lookup_table_pos
```

## Root cause

`unicode_lookup_table_len` is computed as `glyph_count / UNICODE_GLYPHS_PER_LOOKUP_TABLE_ENTRY`
(integer division). When the glyph count is not an exact multiple of 101, the
last (partial) block still needs a table entry, but the length calculation
under-allocates by one. The pending-block code only wrote **one** `0xffff`
sentinel entry, so the final table entry was left unfilled and the assert
fired.

## Fix

- Allocate one extra lookup table entry: `+1` on the length.
- Fill **every** remaining entry with the `0xffff` sentinel (`if` → `while`).

## Validation

Generated a full-CJK font (21058 Unicode glyphs, 21270 total) successfully
with `-b 1`. The output was verified byte-for-byte against the reference
rendering (u8g2 C library draws the same text identically).

## Notes

Requires `U8G2_USE_LARGE_FONTS` (auto-defined on unix/arm) when compiling
the generated font on desktop targets.

### DIFF
--- a/tools/font/bdfconv/bdf_rle.c
+++ b/tools/font/bdfconv/bdf_rle.c
@@ -804,7 +804,7 @@ void bf_RLECompressAllGlyphs(bf_t *bf)
   1 May 2018: Unicode lookup table 
   */
   bf_Log(bf, "RLE Compress: ASCII gylphs=%d, Unicode glyphs=%d", ascii_glyphs, bf->selected_glyphs-ascii_glyphs);
-  unicode_lookup_table_len = (bf->selected_glyphs-ascii_glyphs) / UNICODE_GLYPHS_PER_LOOKUP_TABLE_ENTRY;
+  unicode_lookup_table_len = (bf->selected_glyphs-ascii_glyphs) / UNICODE_GLYPHS_PER_LOOKUP_TABLE_ENTRY + 1;
   //if ( unicode_lookup_table_len > 1 )		
   //  unicode_lookup_table_len--;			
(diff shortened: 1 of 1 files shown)


createskyblue:
The submission is a bit too hasty and is being reviewed...

createskyblue:
Closing this PR — my premise was wrong.

I filed this based on an assertion I hit while generating a full-CJK font, but after rigorous re-testing I found the assertion only occurs in an older bundled bdfconv binary (UNICODE_GLYPHS_PER_LOOKUP_TABLE_ENTRY=100). The current source on master (101) does NOT reproduce it: full-CJK (27775 glyphs), small and sparse maps all generate fine and render all glyphs correctly.

The lookup-table sentinel is intentional (pre-written with encoding 0xffff, kept by the `|=` when the last block entry is written), and my change would have altered the table structure of fonts that already work correctly. Apologies for the noise.

createskyblue:
Withdrawing: not a bug in current master.

olikraus:
thanks for the documentation and (self) review.

PR #2801, opened 2026-08-07
Add driver for MELT MT-12864K (128x64, I2C)

nickgagin:
Add support for the MELT (МЭЛТ) MT-12864K, a 128x64 monochrome graphic LCD module with I2C interface.

## Hardware

The module carries two RW1065 controller/driver chips (KS0108/KS108 compatible command set, controlled in pairs) and one INF8574A (PCF8574A compatible) control register. All three devices are I2C slaves, write only:

- left controller (chip 1): 0x3C
- right controller (chip 2): 0x3D
- control register (INF8574A): 0x3B

The INF8574A register drives `^RST` (reset), backlight and contrast. Reset is done over I2C through this register.

## Implementation notes

- The controllers use the two byte control protocol with a significant C0 bit (datasheet table 5), compatible with `u8x8_cad_ssd13xx_i2c`.
- `tile_width = 16`; the driver splits the tile draw between the left (x < 8) and right (x >= 8) controllers, re-starting the column address for the right half.
- Init sequence (per Amperka/Amp.ru sample code): EXT=1, PS=0, EXT=0, start line 0, sent to both controllers, then Display On.
- The controllers cannot mirror rows/cols and have no internal contrast command (contrast is set via the INF8574A register), so those messages are not implemented.

## Verification

- `sys/sdl/mt12864k_i2c`: simulation test, all transactions verified against the expected byte stream (`ALL TESTS PASSED`).
- Tested on real hardware with the full buffer HW I2C and SW I2C Arduino classes.

### DIFF
--- a/cppsrc/U8g2lib.h
+++ b/cppsrc/U8g2lib.h
@@ -26183,6 +26183,60 @@ class U8G2_KS0108_ERM19264_F : public U8G2 {
     u8x8_SetPin_KS0108(getU8x8(), d0, d1, d2, d3, d4, d5, d6, d7, enable, dc, cs0, cs1, cs2, reset);
   }
 };
+class U8G2_MT12864K_128X64_1_SW_I2C : public U8G2 {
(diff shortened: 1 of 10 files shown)


olikraus:
looks good ....
thanks for the PR

PR #2802, opened 2026-08-12
Prevent dimension underflow in DrawRBox() and DrawRFrame()

Lawrence-Link:
## Description

Prevent unsigned integer underflow in `u8g2_DrawRBox()` and `u8g2_DrawRFrame()` when the corner radius is larger than half of the box width or height.

For example, with `h = 7` and `r = 4`, the existing calculation:

```c
hh = h;
hh -= r;
hh -= r;
```
wraps around because u8g2_uint_t is unsigned. The resulting large length can reach the low-level line drawing function and cause an out-of-bounds framebuffer write. This change returns early when the supplied radius cannot fit within the requested dimensions.

## Testing
I reproduced the issue with an SSD1306 128x64 full-buffer setup, where drawing an animated rounded frame caused a framebuffer access crash.
I rebuilt and tested the application with this change, and the crash no
longer occurs.

### DIFF
--- a/csrc/u8g2_box.c
+++ b/csrc/u8g2_box.c
@@ -87,6 +87,10 @@ void u8g2_DrawFrame(u8g2_t *u8g2, u8g2_uint_t x, u8g2_uint_t y, u8g2_uint_t w, u
 
 void u8g2_DrawRBox(u8g2_t *u8g2, u8g2_uint_t x, u8g2_uint_t y, u8g2_uint_t w, u8g2_uint_t h, u8g2_uint_t r)
 {
+
+  if ( r > (w/2) || r > (h/2) )
+    return;
+
   u8g2_uint_t xl, yu;
   u8g2_uint_t yl, xr;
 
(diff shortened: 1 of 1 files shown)


Lawrence-Link:
After thinking about this a bit more, I realized that returning early and making the entire drawing operation disappear is probably not the most appropriate behavior.

Instead, I now clamp the radius to half of the smaller dimension. When an oversized radius is supplied, the shape gracefully becomes a capsule rather than disappearing. This adds negligible overhead while preserving useful rendering behavior.

I believe an out-of-bounds framebuffer write is inappropriate regardless of the supplied geometry, so the function should handle an oversized radius safely.

olikraus:
Thanks for the PR, I agree to the problem, but adding extra code would increase the code size and execution time.
On the other side this is probably a user mistake and required only for debugging.
More over it is clearly mentioned that the results are undefined if the given conditions are not met: https://github.com/olikraus/u8g2/wiki/u8g2reference#drawrbox

